Issue #3, opened 2016-03-10
Error: Unknown signal: 143


alessioalex:
Hey there!

Go this error:

```bash
→ ./cmd.js examples/rest-api/index.js 
Caught SIGINT, generating flamegraph /home/alessio/www/test/0x/index.js:188
      throw e
      ^

Error: Unknown signal: 143
    at process.ChildProcess.kill (internal/child_process.js:361:11)
    at emitOne (events.js:83:20)
    at process.emit (events.js:170:7)
    at process.exit (node.js:763:17)
    at ChildProcess.<anonymous> (/home/alessio/www/test/0x/index.js:169:15)
    at emitTwo (events.js:88:13)
    at ChildProcess.emit (events.js:173:7)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:201:12)
```

So [`143` corresponds to `SIGTERM`](http://stackoverflow.com/questions/4192364/always-app-java-end-with-exit-143-ubuntu), which is sent upon killing the process. So doesn't that mean that [this line](https://github.com/davidmarkclements/0x/blob/master/index.js#L167) should check for codes `0` as well as `143`? If done so, everything works fine on my system. The generated flamegraph is totally fine.

davidmarkclements:
Sounds good - but just so I can understand - what circumstances are causing the sigterm? Are you sending a kill signal to your process? What OS, and which node version?

alessioalex:
```
 alessio: /home/alessio/www/test/0x git:master
→ uname -a
Linux DXBU-JKSDJ32 3.19.0-15-generic #15-Ubuntu SMP Thu Apr 16 23:32:37 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

 alessio: /home/alessio/www/test/0x git:master
→ node -v && npm -v
v5.2.0
3.3.12
```

I'm just doing `CTRL+C` in order to generate my flamegraph, and I assume `0x` is killing the perf process?

davidmarkclements:
Ahhhh okay yes I see

So on os x the node process and the tracer both started by 0x - we send a sigint to the node process itself and it exits with a code 0

On Linux, perf has to start the node process (if you attach later it fails to get js stacks :() - so we send a SIGINT to perf which sends a SIGTERM to your nose process which seems to result in 143. 

Interestengly it resulted in a 0 (and possibly a SIGINT) on an Ubuntu ami so possibly the version of perf affects how it kills the child (maybe later versions pass the signal through to the child). 

[...]

PR #5, opened 2016-03-10
Closes #2, #3

### LINKED ISSUE #2: Workload failed: No such file or directory
Hey there,

I was trying to run `0x` with the basic Node server example but it's always crashing with this message early on.

Searching Google I found out that this was because of `perf` getting a bad command, but I'm no expert in that.

The command generated by `0x` looks like this on my system (Linux && Node 5.2.x I think):

```bash
sudo perf record -q -e cpu-clock -F 1000 -g -o /tmp/perf-dv1dih.data -- node --perf-basic-prof -r /home/alessio/www/test/node_modules/0x/soft-exit server.js
```

Any tips on how to fix this?

Thank you

alessioalex:


davidmarkclements:
v2.0.6 released - thanks @alessioalex

PR #8, opened 2016-03-14
Fixing a couple of minor problems

unlucio:
- Fixing a problem with browserify shims if the traced project
uses browserify its self
- Adding timestamp to profile dir to avoid collision (ie: in docker
containers)
- Adding Docker trick in readme

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,16 @@ module.exports = function (args) {
   })
 }
 
+function getProfileFolderName(args, proc) {
+  var name = 'profile-' + proc.pid
+
+  if (args['timestamp-profiles']) {
+    name += '-' + Date.now()
+  }
+
+  return name
+}
+
 function sun (args, sudo) {
   var dtrace = pathTo('dtrace')
   var profile = path.join(__dirname, 'node_modules', '.bin', 'profile_1ms.d')
@@ -62,7 +72,7 @@ function sun (args, sudo) {
 
     if (traceInfo) { prof.stderr.pipe(process.stderr) }
 
-    folder = 'profile-' + proc.pid
+    folder = getProfileFolderName(args, proc)
     fs.mkdirSync(process.cwd() + '/' + folder)
 
     pump(
@@ -172,7 +182,7 @@ function linux (args, sudo) {
     }
   })
 
-  var folder = 'profile-' + proc.pid + '-' + (new Date()).getTime();
+  var folder = getProfileFolderName(args, proc)
   fs.mkdirSync(process.cwd() + '/' + folder)
 
   setTimeout(log, delay || 100, 'Profiling')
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -172,7 +172,7 @@ function linux (args, sudo) {
     }
   })
 
-  var folder = 'profile-' + proc.pid
+  var folder = 'profile-' + proc.pid + '-' + (new Date()).getTime();
```
davidmarkclements: Date.now() would be preferred here 

Let's talk about putting this behind an option

package.json
```diff
@@ -39,5 +39,6 @@
   "devDependencies": {
     "deep-diff": "^0.3.3",
     "standard": "^6.0.7"
-  }
+  },
+  "browserify-shim": {}
```
davidmarkclements: nice.

readme.md
```diff
@@ -254,6 +254,21 @@ Yes please!
 
 `DEBUG=0x* 0x my-app.js`
 
+## Docker
+In a docker container, due to security reasons, you most likelly gonna end up with the follwing error:
+
```
davidmarkclements: 1) let's move this up to just beneath "Generating"
2) can we keep the tone consistent (e.g. we, instead of you, avoid colloqualisms) with the rest of the readme, also - typos: 
Due to security reasons Docker containers tend to result in the following error: 

We can work around this problem by running our container with the `--privileged` option, or add `privileged: true` in our `docker-compose.yml` file. See the [Docker docs](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities) for more info

davidmarkclements:
hey @unlucio really appreciate your work here, would you mind going over the inline comments 

Also I think the timestamping is relevant to particular edge cases, what do you think to putting it behind a flag, say `--timestamp-profiles`

unlucio:
@davidmarkclements done, wdyt

usage.txt
```diff
@@ -3,46 +3,48 @@
...
 
-  --cmd | -c      Command Mode - A tool set for processing data in 
-                  the profile folder. 
+  --cmd | -c              Command Mode - A tool set for processing data in 
+                          the profile folder. 
 
-                  Run 0x --cmd help for a list of commands
+  --timestamp-profiles    Add timestamp to you profile folder's name
+
+                          Run 0x --cmd help for a list of commands
```
davidmarkclements: run 0x --cmd should be part of the --cmd flag desc

index.js
```diff
@@ -28,6 +28,16 @@ module.exports = function (args) {
   })
 }
 
+function getProfileFolderName(args, proc) {
+  var name = 'profile-' + proc.pid
+
+  if (args['timestamp-profiles']) {
+    name += '-' + (new Date()).getTime()
```
davidmarkclements: Date.now() is the same thing and less noisy

davidmarkclements:
I think you're awesome - just two tiny niggles - see notes

unlucio:
@davidmarkclements thanks, there you go

davidmarkclements:
pubished 2.1.0 thanks @unlucio

unlucio:
thank you

PR #12, opened 2016-03-20
tooling: adds simpler example, .gitignore: exclude output

eljefedelrodeodeljefe:
Two minor nits:
* it was a little easier to have (tracing unrelated) testing with a simpler example. This commit adds one.
* also IDE experience was a little nicer when output was excluded

I can separate those of course, but they seemed too small at the time.


davidmarkclements:
Beautiful thanks!

PR #13, opened 2016-03-20
feature: adds support for ext. node binaries like ../node/node

### LINKED ISSUE #10: Add support for custom binaries
First of all: (In general) cool work. Always had something like this in mind, but never went for the long distance. Much appreciated.

***

Currently, by looking at the implementation, custom binaries, like when you'd work on node core, don't work. I'd like to be able to run `0x ./node example.js` for benchmarking purposes, whereas `./node` resolves to the node binary in my current working directory.

If you know where to include this real quick, I'd appreciate the change. Otherwise I'll send in a PR at the beginning of the week

eljefedelrodeodeljefe:
Adds actual support for external / custom binaries.

See discussion on #10

re-opens #10, closes #10

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+const path = require('path')
 
 cmd()
 
@@ -20,7 +21,16 @@ function cmd () {
     argv.splice(stacksOnlyIx + 1, 1)
   }
 
-  var ix = argv.indexOf('node')
+  // get the position of node binary in argv; else stay -1
+  var ix = -1
+  var bin = ''
+  argv.forEach((el, index) => {
+    if (path.basename(el) === 'node' && el !== 'node') {
+      ix = index
+      bin = argv[index]
+      return
+    }
+  })
 
   if (ix === -1) {
     var c = argv.length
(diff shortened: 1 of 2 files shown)


cmd.js
```diff
@@ -20,7 +21,16 @@ function cmd () {
     argv.splice(stacksOnlyIx + 1, 1)
   }
 
-  var ix = argv.indexOf('node')
+  // get the position of node binary in argv; else stay -1
+  let ix = -1
+  let bin = ''
```
davidmarkclements: since 0.12 can emit perf events, I'm reluctant to use syntax that isn't supported in 0.12 (even though I've not tested in 0.12) 

so can we make let, var 

we can leave const, since 0.12 does support that (even though it's not the es6 const, but the old v8 const)

davidmarkclements:
sorry just one thing then merge :D

eljefedelrodeodeljefe:
no worries. Also had a typo in there. Force pushed.

davidmarkclements:
yah even though it won't on .12 (because of -r flag) there's still the possibility of supporting it (e.g. by requiring it in a parent file, and then using hacks like Module.runMain() so it thinks it's the top module - as in https://github.com/davidmarkclements/nodux-core/blob/master/include/usr/local/bin/nodux.js) 

tested it all - works great trying to publish but as I say crappy 3g connection - 2.2.0 will be out as soon as I can pub. (minor bump because 0x ./node is technically an api addition)

eljefedelrodeodeljefe:
Yeah, looks good to me. as a matter of opinion I usually tend to ignore pre 4.X versions for my node stuff :) . But obviously, if easily possible - why not.

davidmarkclements:
added #14 - if there's demand I'll implement - if not then no point

davidmarkclements:
hooray 2.2.0 published: https://www.npmjs.com/package/0x

PR #15, opened 2016-03-21
feature: adds autoexit option. Default: do autoexit after script

eljefedelrodeodeljefe:
@davidmarkclements maybe this would be interesting for you. Don't know what your plan for this was, but I like the notion of autoexiting, after the script has run. It will however wind down the rest of the app logic.

### DIFF
--- a/examples/simple/index.js
+++ b/examples/simple/index.js
@@ -1,17 +1,18 @@
 'use strict'
-let count = 0, max = 10, res = 0
+var count = 0
+var max = 10
+var res = 0
 
-console.log('Starting process with simple example.');
-setTimeout(()=>{
-  setInterval(()=>{
+console.log(`Starting process(${process.pid}) with simple example.`)
+setTimeout(() => {
+  setInterval(() => {
     res += 1
     count++
-    console.log(res);
+    console.log(res)
 
     if (count >= max) {
-      console.log('Process exited now. Hit ctrl+c');
+      console.log('Process exited now. Hit ctrl+c')
       return process.exit()
     }
-
   }, 10)
 }, 500)
--- a/index.js
+++ b/index.js
@@ -67,7 +67,14 @@ function sun (args, sudo, binary) {
   })
   var folder
   var prof
-  function start() {
+
+  function start () {
+    // on script end, bail automatically, don't when no-autoexit flag is set
+    proc.on('exit', () => {
+      debug(proc.pid, 'exiting')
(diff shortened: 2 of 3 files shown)


examples/simple/index.js
```diff
@@ -1,17 +1,18 @@
 'use strict'
-let count = 0, max = 10, res = 0
+let count = 0
+let max = 10
+let res = 0
```
davidmarkclements: let -> var

davidmarkclements:
Yah this is great, I don't think we need the option, just do it on exit, otherwise why use 0x in the firstlace

examples/simple/index.js
eljefedelrodeodeljefe: will do.

eljefedelrodeodeljefe:
Then I'll remove the option.

eljefedelrodeodeljefe:
Done.

davidmarkclements:
Outstanding thanks man

davidmarkclements:
just need to check/get linux behaviour the same on this

eljefedelrodeodeljefe:
Sure. I have parallels installed. Will check too.

eljefedelrodeodeljefe:
Apologies if I missed it.

eljefedelrodeodeljefe:
I know where the issues lie. See comments. basically I was mistaken the start function to be global. whereas it should be fine send `process.kill(proccess.pid, 'SIGINT')` in the exit event of the spawn functions directly. I'll adda global function and call at in `sun()` and `linux()` respectively.

davidmarkclements:
yah cool - by global you mean scoped of course - not actually adding to `global`

eljefedelrodeodeljefe:
Call it [here](https://github.com/davidmarkclements/0x/pull/15/files#diff-168726dbe96b3ce427e7fedce31bb0bcR66) and [here](https://github.com/davidmarkclements/0x/pull/15/files#diff-168726dbe96b3ce427e7fedce31bb0bcR189)

eljefedelrodeodeljefe:
Yes. Scope in the file. Actually extra function won't be needed. I'll send in the thing later.

eljefedelrodeodeljefe:
This shows it's behaviour https://www.youtube.com/edit?o=U&video_id=QzQa34qZKOU

davidmarkclements:
link is broken (maybe just visible to you if logged in)

eljefedelrodeodeljefe:
This one, sorry https://youtu.be/QzQa34qZKOU

PR #18, opened 2016-03-22
feature: adds autoexit option. Default: do autoexit after script

eljefedelrodeodeljefe:
Addresses Revert #17 of #15. 

Note: there might be unrelated race conditions and an unrelated bug on linux, when called with binary or plain `node` binary argument.

PR #19, opened 2016-03-22
fix: potential race conditions during event dispatch; refactor

eljefedelrodeodeljefe:
This PR mainly adds the event emitter to dispatch states. This was necessary, since their have been potential race conditions around the dispatch of SIGINT.
Also, and more importantly the motivation for this, after introducing the auto exiting mode (I think) you haven't been able anymore to exit the target program early through SIGINT and still get profiling results. Maybe it didn't work before also. Either way this is possible again.

Since the listeners needed to physically move, I had to move variable declarations, too, and hope you don't mind the refactoring.

During refactor there also was a problem with the "leaking clock", which is why I introduced central SIGINT and exit handling at the top of the program, and destroying allocated memory for it. 

Seemingly getting all the shared variables in the main scope screams for refactoring this as class. I of course refrained from doing so, but wanted to discuss this (maybe)

**Note:** this hasn't altered the API

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,27 @@ var convert = require('./stack-convert')
 var gen = require('./gen')
 var debug = require('debug')('0x')
 var log = require('single-line-log').stdout
+var EventEmitter = require('events')
+var ee = new EventEmitter()
+
+var traceInfo
+var profile
+var proc
+var prof
+var folder
+
+var clock
+// handle everything that might be leaking here
+process.on('exit', () => {
+  if (clock) {
+    clock.kill()
+  }
+})
+// handle premature exit here (when already registered .on('target_exited')).
+// REALLY early exiting (right after start) is handled in listener
+process.on('SIGINT', () => {
+  ee.emit('target_exited')
+})
 
 module.exports = function (args, binary) {
   isSudo(function (sudo) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -232,11 +225,42 @@ function linux (args, sudo, binary) {
         sink(args, proc.pid, folder, clock)
       )
     })
+  })
 
+  proc = spawn('sudo', [
```
eljefedelrodeodeljefe: @davidmarkclements please note, that the diff looks larger then it is, because I basically just reordered code blocks. For git that's a deletion essentially it's just moving around things though.

eljefedelrodeodeljefe: Covered in it though is the change of the variable declarations that moved to the top.

davidmarkclements:
@eljefedelrodeodeljefe thanks for all your efforts man, really appreciate your input

can you speak more to the problems here,  (I get the clock issue, what are the other race conditions?) 

I'm not keen on this approach - it seems that we may be adding unnecessary complexity, and I'm particularly against the "exoskeletal" concept of shared variables (and esp. against classes). 

I'd rather have duplicated code than go down this road - can we try to fix this another way?

davidmarkclements:
so I had to a chance to look into this, the race conditions and errors were caused in #18 - 
the problem was that calling process.kill(process.pid, 'SIGINT') caused a circular dependency of sorts, 
the sigint handler tries to kill the child process, whilst when the child process exits it triggers the sigint handler.

eljefedelrodeodeljefe:
Okay that potentially fixes it too. Didn't see any problem at least. Will investigate more.

Issue #30, opened 2016-04-15
electron wrapper


davidmarkclements:
display flamegraph.html in electron

1. should this be default behaviour (of a v2)
2. should it be behind a flag
3. if behind a flag, should we make users download electron etc. during default install if they don't intend to use this feature (e.g. running 0x on server side)

mcdonnelldean:
If not picked up by Monday let me know I can get someone on it

davidmarkclements:
@eljefedelrodeodeljefe can have first dibs on this - if he wants it - since he's made a lot of great contributions already - but great to know help is there - no harm if we split the task into pieces either

on that here's how I think we should do this, any input appreciated

1. have the ability to automatically "open with" any specified command, e.g. the `open` command on OSX  so it automatically opens in the default browser
2. use this same ability to open with electron, possibly overlaw a convenience flag for this special case
3. when opening with electron, we mean opening with the electron app that enhances the flame graph interaction experience - primarily by allowing file system access via the stacks - so possibly this app that enhances the experience should be a separate tool, that's separately installed

@mcollina @mafintosh @eljefedelrodeodeljefe @mcdonnelldean - what do you think?

mafintosh:
although the "open with" idea sounds nice and simple a "pure" electron integration would mean that it wouldn't need to write any files to disk (assuming that is even possible to configure) which is nice if you a bunch of profiles while testing stuff out

mafintosh:
so :+1: on 3.

mcdonnelldean:
I'm good with all that 👌 Oooh can't wait to play with this when done! Yaaay

mafintosh:
Crazy idea: it could even have a mode where it re-ran benchmarks and updated the flamegraph when underlying deps changed. Also side-by-side comparison and stuff with older versions.

davidmarkclements:
@mafintosh good point so we've got two scenarios for an electron app here

1. instant viewer, no trace of files on system - put them in /tmp if we need them - just make the experience seamless
2. viewer of profiles that we have kept, e.g. cases where we want the profile folder, or cases where we want to view a profile generated some time in the past 

it's interesting you should say that, because @mcollina and I were discussing this very thing, along with branch comparisons etc., and integrating http://npm.im/autocannon for http benchmarks etc. - we we're calling it the Performance Toolkit.

This could be the start of that, which is another reason I'm suggesting as a separate app that integrates

davidmarkclements:
so here's the current order of events:

```
[...]

PR #31, opened 2016-04-21
feature: electron, initial stab

eljefedelrodeodeljefe:
See https://www.youtube.com/watch?v=F7zGclpMXkA

Handles also people not wanting to install electron by `NO_OPEN_0X=true` environment variable and conditional, programmatic `npm install` via `preinstall` hook.

Was a quick one, but I think it is extensible.


mcollina:
So I think we can make the process interactive: the user launch 0x, and if it _thinks_ it should open electron, it will ask the end user and then it will start the download of prebuilt. I would use the same process https://www.npmjs.com/package/insight uses to remember the choice.

We can possibly detect if we can open a desktop app by looking at some env variable.

BTW, we should add insight as well.

davidmarkclements:
@eljefedelrodeodeljefe nice touch with the icon 😋 

@mcollina @eljefedelrodeodeljefe I'm not sure this should be part of 0x (as per https://github.com/davidmarkclements/0x/issues/30#issuecomment-211340639) but rather a separate module - this side-steps the whole download prebuilt if config specifies to use 0x

at any rate we can always pull it out if need be and work backwards with the `open` flag

next steps is to integrate file viewing - I'm thinking we pass a "view" opt, which is a function that gets called when the user (double clicks? cmd clicks? 1st click widens, 2nd click views?) on a frame

It would be real nice if the user could associate a view command - e.g. `vi` of `subl` etc. etc. could default to `$EDITOR` env var.

At the same time, a preview popout box could open (on hover?) at the particular line, with 2-3 surrounding lines above and beneath 

ref: #30

eljefedelrodeodeljefe:
Hmm okay. I'll continue, but it's going to be harder with the option passing in to another module. But why not. 

I am already including the CLI prompt as @mcollina suggested. That would probably need to stay in this module as well as well as the option persisting.

I'll send in new stuff later.

eljefedelrodeodeljefe:
@davidmarkclements this turns out to be a bigger problem. We would need the process.kill / process.on('exit') refactoring I was proposing unfortunately. Process.on('exit') allows only synchronous function calls in the handler. Most of the interactive logic (and in the feature other probably as well) is async. 

Can you decide on this quickly so I can do it right away? Otherwise we cannot have the conditional install logic in the app.

davidmarkclements:
hey @eljefedelrodeodeljefe sorry for not getting back sooner - 

I think the approach as it stands is too coupled - what I'd rather see is an independent command being executed on process.exit (btw, love your thinking on passing path through process.exit, never would have thought of that) 

So 0x-view should have it's own bin - rather than be required in.

Advantages are

1. if we decide to go with the conditional install or prebuilt, the separate bin can handle that
2. the bin command can be easily configured via an open flag (e.g. you could set it to chrome etc.)
3. it makes it v. easy to extract into it's own module later

We also need to take out the NO_0X_VIEW thing out, first I'm not sure 0x-view should be default, and at any rate it doesn't cater to people who are including 0x via package deps, potentially as part of a build process (like @yoshuawuyts)

davidmarkclements:
hey @eljefedelrodeodeljefe we can talk more about this offline - I think this should be a separate project

eljefedelrodeodeljefe:
Yep. Like the idea.

Issue #34, opened 2016-06-13
Error with generating flamegraphs when starting server using babel require hook


craigkj:
Hi there,

There appears to be a problem with generating the flamegraphs after profiling a node app using the babel require hook (https://babeljs.io/docs/setup/#installation).

```javascript
require('babel-register');
// Setup my server and start it
```

After terminating the app a `Unable to find map file!` and the waiting clock hangs indefinitely. This is identical to the bug found here: https://github.com/davidmarkclements/0x/issues/33 which I debugged a little further to try and pin down the problem.

The following note from the babel documentation possibly indicates the problem:

```
The require hook automatically hooks itself into all node requires. This will pollute the global scope and introduce conflicts. Because of this it's not suitable for libraries, if however you're writing an application then it's completely fine to use.
```

davidmarkclements:
Yah we need to figure out an elegant way to co-exist, or at least a user friendly work around

Needs research, cc @mcollina @lucamaraschi @eljefedelrodeodeljefe

craigkj:
@davidmarkclements

One possible work around is to pre transpile the app using the babel CLI and then remove the reference to the require hook for the duration of the profiling. Using the require hook in production is now (fairly newly) recommended against so I think a work around is fine.

The above work around worked for me and i'm now getting the flamegraphs generated as expected, which is glorious.

I'm happy to compile an example and some documentation/notes and PR them to the app if you would like?

davidmarkclements:
@craigkj that would be very much welcome and appreciated - thanks for offering :bowtie:

davidmarkclements:
covered by #35

Issue #33, opened 2016-06-07
`Unable to find map file!` when running against Express server


Limess:
I get the following output and a hanging clockface icon when I try to use `0x` on OSX 10.10.5. 
I'm running `0x` on an express application using node v4.4.4.

When I run 
`DEBUG=0x* 0x -p --trace-info node server.js`
Upon sending a `SIGINT` (`ctrl+c`) to `0x` I get the following message: 
```
0x Caught SIGINT, generating flamegraph +0ms
Unable to find map file!
  0x unable to find map file +23ms
  0x tidying up +12ms
```

My terminal then has an instance of `/Users/myuser/.nvm/versions/node/v4.4.4/lib/node_modules/0x/node_modules/.bin/clockface` hanging on the prompt.

A `profile-${number}` folder is created with a `.stacks.${number}.out` file within, however I can't generate a flamegraph from this, receiving the following errors on separate projects
```
(function (exports, require, module, __filename, __dirname) { node 8169 494393: profile-1ms:
                                                                   ^^^^

SyntaxError: Unexpected number
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at startup (node.js:139:18)
    at node.js:968:3

```
and

```
/Users/myuser/.nvm/versions/node/v4.4.4/lib/node_modules/0x/stack-convert.js:31
  var res = {
  ^

RangeError: Maximum call stack size exceeded
    at Node.serialize (/Users/briggc08/.nvm/versions/node/v4.4.4/lib/node_modules/0x/stack-convert.js:31:3)
    at Node.serialize (/Users/briggc08/.nvm/versions/node/v4.4.4/lib/node_modules/0x/stack-convert.js:40:38)
    at Node.serialize (/Users/briggc08/.nvm/versions/node/v4.4.4/lib/node_modules/0x/stack-convert.js:40:38)
    at Node.serialize (/Users/briggc08/.nvm/versions/node/v4.4.4/lib/node_modules/0x/stack-convert.js:40:38)
    at Node.serialize (/Users/briggc08/.nvm/versions/node/v4.4.4/lib/node_modules/0x/stack-convert.js:40:38)
[...]

PR #35, opened 2016-06-15
Add basic babel example and notes on babel require hook incompatibility

craigkj:
## Changes
- Add example babel project using restify
- Update readme with notes about babel require hook incompatibility 

## Background

Added as a response to the issues: https://github.com/davidmarkclements/0x/issues/33 & https://github.com/davidmarkclements/0x/issues/34 

0x currently has some unknown issues when run against a project using the babel require hook (https://babeljs.io/docs/setup/#installation). The error you get when doing this is not ideal so hopefully these additions should save someone else some pain should they also encounter it!

This attempts provide an example of a work around where 0x is run against a babel project using the babel-cli instead of the require hook. Notes informing users of the incompatibility of the require hook have been added to the example readme and the base project readme. 

The example contains a `src` folder for your es6 code, and a `lib` folder which is where babel puts the transpiled es5 code when you run the transpilation command: `npm run build`.


davidmarkclements:
Outstanding job @craigkj thanks so much for this, I'll take a closer look in the morning then merge & release

craigkj:
@davidmarkclements no problem, shout if you have any questions/concerns/improvements. Main gola is just to highlight problem to save others some pain :D

PR #55, opened 2016-12-01
Detect location of profile_1ms.d as .bin resident

### LINKED ISSUE #54: Running on OSX with NPM3 does not work.
profile_1ms.d is being referenced absolutely in the directory structure for NPM2. In NPM3, it is placed in the node_modules/.bin directory, rather than the previous home of node_modules/0x/.bin directory.

https://github.com/davidmarkclements/0x/blob/master/index.js#L48

I will try to toss together a PR.

@davidmarkclements - Thanks for the on-spot debugging at the conference yesterday. Turns out I had two issues, XCode and then this issue.

lostthetrail:
Resolves #54

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ function getProfileFolderName(args, proc) {
 
 function sun (args, sudo, binary) {
   var dtrace = pathTo('dtrace')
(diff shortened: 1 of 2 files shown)


davidmarkclements:
thanks @lostthetrail !

PR #61, opened 2017-01-20
feat: support custom output directory via -o / --output-dir

### LINKED ISSUE #60: Support specifying output directory
It'd be nice if it were possible to specify the output directory to use instead of [generating it on the fly using the `PID`](https://github.com/davidmarkclements/0x/blob/4074aa2021adaef208d9fc9fd239e8e68f38df3a/index.js#L37).

Especially in a docker world where specific files and directories are mounted into containers, having non-deterministic paths can be a pain.

Something like `-O / --output-dir` would be awesome 🙂

Thanks for putting out such a quality tool!

knksmith57:
resolves #60

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,13 +34,29 @@ module.exports = function (args, binary) {
 }
 
 function getProfileFolderName(args, proc) {
-  var name = 'profile-' + proc.pid
+  var name = args['output-dir']
+
+  if (!name) {
+    name = 'profile-' + proc.pid
+  }
 
   if (args['timestamp-profiles']) {
     name += '-' + Date.now()
   }
 
-  return name
+  return path.resolve(process.cwd(), name)
+}
+
+function ensureDirExists(path) {
+  try {
+    fs.accessSync(path)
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -34,13 +34,29 @@ module.exports = function (args, binary) {
...
 
   if (args['timestamp-profiles']) {
     name += '-' + Date.now()
   }
 
-  return name
+  return path.resolve(process.cwd(), name)
+}
+
+function makeDirectoryIfNotExistsSync(path) {
```
davidmarkclements: `ensureDirExists`

or 

`mkdirp`

davidmarkclements:
Hey thanks for the PR! This is a useful feature, just a small change req and then we can merge 👌

index.js
davidmarkclements: Actually the former since mkdirp is really about creating nested structures

knksmith57:
Awesome, thanks @davidmarkclements. PR feedback addressed

PR #64, opened 2017-03-15
Include troubleshooting remedy for empty stacks

blake-regalia:
From personal experience.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -132,6 +132,13 @@ For Chrome, run with the following flag
 
 Where 8024 is the megabytes of RAM required to run load stack. Adjust this as needed and confirm you have it to spare.
 
+## Empty Output Stacks
+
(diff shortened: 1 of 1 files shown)


mcollina:
@blake-regalia which combination of operating system/node version/install method required you to do this?

It would be good to note this down in the README. We are already escalating the privileges for the thing needed, so if we need to change some permissions, we can call `sudo`  from within 0x.

blake-regalia:
I have npm install global packages under my home directory, so that they don't have root privileges by default.

```sh
$ uname -msr
Darwin 13.4.0 x86_64

$ npm config get prefix
~/Users/blake/.npm-packages

$ node -v
v7.7.2
```

mcollina:
Ooh, that's an oldish version of Mac OS X. That might very well be then.

PR #67, opened 2017-04-01
Killing the profiling process if it stubbornly refuses to die

### LINKED ISSUE #66: Mac: Prof process does not die
In some of my executions, after the child process exits, the profiling process does not die.
One interim solution was to send a SIGKILL instead of a SIGINT to the profiling process [here](https://github.com/davidmarkclements/0x/blob/master/index.js#L152).
Have you seen this happening before, and if so, did you get a sense of why this may happen?

Node v6.9.2
MacOS 10.12.3.

Thanks so much, great module BTW!

pgte:
Also, fixed max attempts at expecting prof process dying.
Should fix #66

### DIFF
--- a/index.js
+++ b/index.js
@@ -155,7 +155,11 @@ function sun (args, sudo, binary) {
     function capture (attempts) {
       if (!profExited) {
         if (attempts) {
-          setTimeout(capture, 300, attempts--)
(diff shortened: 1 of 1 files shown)


mcollina:
Thanks!

PR #78, opened 2017-10-03
Some README and cmd fixes, added --node-options

mcollina:


davidmarkclements:
can we see `--node-options` example with multiple flags and also check that this works

mcollina:
Done

Issue #70, opened 2017-06-13
Passing arguments to the script being executed fails when `node` not specified


edmorley:
Using 0x 2.4.2, nodejs 8.1.0 under Ubuntu 16.04, passing options to the script being run fails with an unclear error message:
```
$ 0x ./node_modules/.bin/oao bootstrap
module.js:487
    throw err;
    ^

Error: Cannot find module '/home/vagrant/neutrino-dev/bootstrap'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:575:3
```

By experimentation I found that the alternative form works however:
```
$ 0x node ./node_modules/.bin/oao bootstrap
...
```

I'm presuming both forms are expected to work?

Many thanks!

mcollina:
@edmorley no, only the second one is supported atm. But if you want to send a PR we'll be very happy to review it!

edmorley:
Ah ok - I thought either were, since the docs say "When node flags aren't required, we can use the shorthand form" -- and I'd interpreted "node flags" to be "nodejs runtime options" rather than "options passed to the script itself".

davidmarkclements:
Fixed in v3

PR #80, opened 2017-10-06
pass command line arguments through to spawned process

rwallace-kabam:
I didn't make the corresponding linux change because I'm not currently set up to test it.  It would be the same simple concatenation.

Also added restify dependency

https://github.com/davidmarkclements/0x/issues/70

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -46,6 +46,7 @@ function cmd () {
   }
 
   args.script = args._[0]
+  args.script_args = args._.slice(1);
 
   require('./')(args, args.node)
 }
--- a/index.js
+++ b/index.js
@@ -79,7 +79,7 @@ function sun (args, sudo, binary) {
   args = Object.assign(args.nodeOptions.concat([
     '--perf-basic-prof',
     '-r', path.join(__dirname, 'soft-exit')
-  ]).concat(args.script), args)
+  ]).concat(args.script).concat(args.script_args), args)
 
   var proc = spawn(node, args, {
     stdio: 'inherit'


rwallace-kabam:
Someone who's well set up to test linux should probably make the corresponding change - just concatenating the script_args in the call to spawn.  I didn't want to tie it up in this PR since I can't test it.

davidmarkclements:
0x ... doesn't depend on restify 

There might be an example folder with restify with its own package.json which should have restify as a dependency - but 0x itself has no need for restify

Regarding script args this used to work, @mcollina did that previous PR that required addition of node_options flag also break script flags?

mcollina:
Possibly! @rwallace-kabam can you please provide an example that is broken?

rwallace-kabam:
@davidmarkclements yes, restify was only required for the example - my bad.

As for the script args, it looks like https://github.com/davidmarkclements/0x/issues/70 describes exactly what I was running into, and i found that his "alternate" solution including node in the call did not help.

To see a broken example, just `console.log(process.argv)` in examples/rest-api/index.js and run `0x examples/rest-api foo bar`.  See that `foo` `bar` are not passed through.

Also, thanks for the great talks and performance workshop at node interactive :)

rwallace-kabam:
I've removed the commit adding restify as a dependency

rwallace-kabam:
Pre-change:
screen shot 2017-10-06 at 5 13 07 pm

Post-change:
screen shot 2017-10-06 at 5 14 21 pm

PR #85, opened 2018-01-18
V3

davidmarkclements:
todo: Readme updates

for command wrapping: `require('0x/cmd')(process.argv.slice(2))`


mcollina:
Can you please flip the switch on `--open` option, so that it opens the flamegraph by default if it is not specified?

usage.txt
```diff
@@ -1,13 +1,18 @@
...
+
+  --name                  The name of the HTML file, without the .html extension
+                          Can be set to - to write HTML to STDOUT
+                          Default: flamegraph
 
   -o                      Automatically open after finishing
                           Default: false
 
-  --node=PATH             Path to the node executable to use
+  --node-path             Path to the node executable to use
```
mcollina: I think this option is completely redundant. Can you remove it?

usage.txt
```diff
@@ -1,13 +1,18 @@
...
                           Default: false
 
-  --node-options=ARGS     List of options to pass to Node
+  --node-options          List of flags to pass to Node, can be used
+                          instead of the double dash suffix (--)
```
mcollina: can you remove this?

mcollina:
The generate title should not include the `0x --argument --another-argument` part, but only the `node server.js` part. I would also filter out all node core arguments if possible.

usage.txt
```diff
@@ -1,13 +1,18 @@
 
-  0x [flags] file.js
+  0x [flags] -- node [nodeFlags] script.js [scriptFlags]
```
mcollina: Can you make the `0x` part in this document customizable? It will help for integration purposes.

mcollina:
Can you store the file as: `12345.flamegraph` instead of `profile-12345`?

LICENSE.md
```diff
@@ -1,4 +1,4 @@
-All code other than flamer.js and stack-convert.js is licensed thusly:
+All code other than ui.js and stack-convert.js is licensed thusly:
 
   The MIT License (MIT)
   Copyright (c) 2016 David Mark Clements
```
mcollina: You should probably update the copyright year.

usage.txt
davidmarkclements: added a customisable banner option

davidmarkclements:
* `0x` launches help
* output folder renamed
* title shows node command instead of 0x command – (core arguments left, I think it's useful if they stay, we can optionally hide in the ui if necessary)
* --node-path removed
* --node-options removed
* license updated
* banner option added (instead of customisable name in usage text, we can output banner as desired)
* regarding `--open`, auto opening in a browser – the whole paradigm is brittle because it's based on system configuration – I also personally don't like that functionality, I prefer to open it myself (cmd + click file URI) – (for our wrapper project we can pass that arg if that's default behaviour across the board) 
* visualize-only and collect-only added
* readme update
* programatic example added

PR #91, opened 2018-01-22
Use opn instead of open for detached opening

### LINKED ISSUE #89: --open does not work on linux
the `open` module isn't detached on linux. that means that when the 0x process dies, the browser tab dies as well. In practice this means that 0x never shows the tab as it explicitly exits itself on when done.

A solution is prob to switch to https://github.com/sindresorhus/opn

mafintosh:
Fixes #89


davidmarkclements:
3.0.1 released

PR #93, opened 2018-01-22
closes #87

### LINKED ISSUE #87: cli: fail with more gentle warning saying that 0x is not supported on windows
as titled

davidmarkclements:


mcollina:
I meant something slightly different. This is currently emitted as an error with a stacktrace. a CLI tool should just write a simple message to stderr and exit with code 1.

davidmarkclements:
merging for now because it also makes platform generic

Issue #83, opened 2017-10-27
Interpreter frames are translated wrong in Node 8


AndreasMadsen:
_@mcollina asked me to share some details about interpreter frames._

When JavaScript functions aren't JIT compiled but interpreted the instruction addresses will point to the interpreter handlers and not compiled JavaScript frames directly. I have received the following information from V8 team member Jaroslav Sevcik on how this can be solved:

> Also, as Benedikt points out, you will have trouble with interpreted functions because the instruction pointer points to the bytecode handler rather than to the actual JavaScript function. To get around this, you might want to use `--log-code` because that will log bytecode addresses, as well. When walking the stack, you could collect the bytecode array address as it lives in a fixed position in `InterpreterEntryTrampoline`'s frames (see the [SafeStackFrameIterator](https://cs.chromium.org/chromium/src/v8/src/frames.h?type=cs&sq=package:chromium&l=1213) class, which is responsible for collecting the stacks for v8's builtin profiler). This could be later resolved using the -`-log-code` log -- for InterpreterEntryTrampoline, you would simply interpret the bytecode array using the `--log-code` log.

davidmarkclements:
so the problem is, correlating the addresses - they don't match in an obvious way – there's possibly variable offsets happening 

on a side note, there's a way to ensure (hopefully almost) everything JIT compiles

```
0x -- node --always-opt app.js
```

davidmarkclements:
I'm starting to think (for the second time in a year) that this is going to require patches to core, 
for dtrace to the src/v8ustack.d file and for perf – I'm not sure where.

How else do we extract the bytecode array from *within* InterpreterEntryTrampoline frames.. 
If I'm understanding correctly (which I ~~might not be~~ quite probably am not) "InterpreterEntryTrampoline frames" as cited is a set piece of memory with offsets, we get those offsets from somewhere (from what I can see we've got 4096 byte chunks - pages, and then dynamic offsets within those). We have to be able to access that memory at tracing time, which means we have to modify the ustack helper (for dtrace, again for perf, no idea yet).

@AndreasMadsen @mcollina @mafintosh can you verify what I'm saying? Am I seeing this right?

davidmarkclements:
yeah. now I remember: https://github.com/davidmarkclements/0x/issues/76#issuecomment-324476489

AndreasMadsen:
For sure you can implement your own ustack helper, but that only solves the issue for SmartOS.

[...]

PR #98, opened 2018-02-06
phase option, closes #90

### LINKED ISSUE #90: Does not work on Linux if a program ends before you kill it
On Linux, try profiling

``` js
for (var i = 0; i < 100; i++) {
  Math.random()
}
```

The program ends before you can `ctrl-c` the process and it dies crashes with

```
kill: sending signal to 15563 failed: No such process
```

consistently. Seems to work on Mac tho.

davidmarkclements:
+@mafintosh 

The `--phase` option does away with default delay of 300ms 

By default it strips away all startup stacks, everything in the first tick (`--phase` is 2)

If `--phase` is 1, then any stacks from Module.runMain (e.g. execution from the entry point file) are included, these still tend to dominate because the applications module tree is loaded from this point but in some cases (say a CLI tool) this would be useful

`--phase` 0 includes everything in the bootstrap process, including all NativeModule stacks etc. 

Also include a programmatic API option `phase` (as the flag) and `mapFrames` which allows for a custom mapping function.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+# v3.2.0
+* introduce `--phase` option
+* change `--delay` from `300` to `0` - not a breaking change 
+because `--phase` provides the same result (stripping module loading stacks) 
+far more reliably.
+
 # v3.1.0
 
 * enhance `--output-dir` option with interpolation feature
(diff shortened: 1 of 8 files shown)


davidmarkclements:
oh also is closes #90 because with delay being set to 0 all stacks are captured

davidmarkclements:
btw see https://github.com/davidmarkclements/0x/issues/83#issuecomment-363214720 for generating all flamegraph with stacks in Node 8

PR #99, opened 2018-02-08
--prof-viz extra flame graph based on v8 prof data 

davidmarkclements:
comments welcome

@mafintosh @bridgeAR @lucamaraschi @mcollina

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 # v3.3.0
 * introduce experimental `--prof-viz` flag, generates additional flamegraph based on internal v8 profiling data
+* linux fixes & tidy up
 
 # v3.2.0
 * introduce `--phase` option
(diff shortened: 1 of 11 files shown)


davidmarkclements:
Yes. I'll do it as a next step

davidmarkclements:
ok here's the plan. 

release --prof-viz as the new minor – no windows support

make prof the default stack source for 0x v4, this will be the first 0x to support Windows. It also means 0x will work by default with zero configuration inside VM's and containers. 
but will be for Node 8+ only (we rely on v8s tick processor, that only really works in Node 8)

v4 will support `--kernel-trace` (and `--kernel-trace-only`, these will continue to use stack 
collection for os x/solaris and linux - which can be useful for capturing I/O stacks (libuv stuff) 

this will lead to the following compatibility map

| 0x | node       | maxOs | linux | windows |
|----|------------|-------|-------|---------|
| v1 | 0.10, 0.12 | ☑️     | ☑️     | ⤬       |
| v2 | v4, v6     | ☑️     | ☑️     | ⤬       |
| v3 | v6, v8+    | ☑️     | ☑️     | ⤬       |
| v4 | v8+        | ☑️     | ☑️     | ☑️       |

PR #105, opened 2018-03-14
V4

davidmarkclements:
4.0.0-rc1 published to npm


davidmarkclements:
4.0.0-rc2 published to npm

mcollina:
schermata 2018-03-14 alle 10 32 28

The bottom bar should have some bottom margin, it risks not being showed when opening the file on lower resolutions.

mcollina:
This is broken on master/node 10.

davidmarkclements:
4.0.0-rc3 published to npm 

response ui, resolves https://github.com/davidmarkclements/0x/pull/105#issuecomment-372958283

davidmarkclements:
not currently working on windows (can't find isolate log file)

davidmarkclements:
regarding master/10.0.0-pre – this is a problem with master branch, the reported v8 version on `process.versions.v8` and the v8 version header output in the isolate log file are different. 

You can prove this is the problem by editing the isolate log file header from

```
v8-version,6,3,292,48,-node.6,0
```

To

```
v8-version,6,2,414,50,node.6,0
```

Before pressing Ctrl+C and then the flamegraph will generate 

Not a blocker for v4 release

davidmarkclements:
v4.0.0-rc5 resolves https://github.com/davidmarkclements/0x/pull/105#issuecomment-373064643 (fixed on windows)

mcollina:
Re https://github.com/davidmarkclements/0x/pull/105#issuecomment-373140821-

@davidmarkclements doesn't the isolate file include the PID? Why do you have to match also on the specific node version?

mcollina:
The current <- flame -> identifier does not make things clearer. Can you change it to "show all", or "zoom out"?

Can you add the `--on-port` flag as it's implemented in doctor? cc @mafintosh (https://github.com/nearform/node-clinic-doctor/blob/3c19ce2ad380d4a56c69677fb1d8f44163a29751/index.js#L44-L50),
`clinic doctor --on-port 'autocannon -c 100 localshot:$PORT' -- node server.js`. So you get a nice one-liner.

schermata 2018-03-16 alle 11 26 01

There is something wrong with the border in the buttons, as there is some white between the color and the border.

davidmarkclements:
I think the   ⇦  🔥 ⇨ is fine?  It's just a reminder that you click blocks to open up stacks, and clicking the base block opens up all stacks (if we we're to change it though, I'd go with "all stacks")

Yes I think we should get `--on-port` in before release - @mafintosh has been asking for this feature in v4 as well

Re the button border - that's because it's an outline, and not a border. An outline is much easier because it won't cause any layout shifting, you don't have to compensate with the margins. 

I don't see that as a problem

mcollina:
> ⇦ 🔥 ⇨

Does not mean anything.

> I don't see that as a problem

It is just ugly.

davidmarkclements:
I tried "all stacks" it does look better

davidmarkclements:
v4.0.0-rc6  

* resolves final UI issues
* visualize only not dependent on node version
* readme and docs

still need to

* integrate --on-port (@mafintosh did you want to integrate? either way I'm happy, just lmk)
* cover a complex edge case to do with inlinable functions

PR #106, opened 2018-03-19
V4 with --on-port

mafintosh:
Works! but need to expose the cli options i think (see examples/api.js)


davidmarkclements:
Awesome!

davidmarkclements:
merging to v4

PR #116, opened 2018-04-04
fix: error correcting perf not found error message

thlorenz:


davidmarkclements:
thank you! v4.0.9 released

Issue #120, opened 2018-04-17
Error: "toString()" failed


davisjam:
I am running 0x on the [Node-DC EIS](https://github.com/Node-DC/Node-DC-EIS/) benchmark.

After Ctrl-C'ing to deliver SIGINT I got this `toString` failure:

```
$ 0x --kernel-tracing --output-html /tmp/eis-flamegraph.html server-cluster.js 
🔥  ProfilingStarting a single instance of Node.js process with (pid: 18802)
********************************
Server running at port: 9000
********************************
(node:18802) DeprecationWarning: `open()` is deprecated in mongoose >= 4.11.0, use `openUri()` instead, or set the `useMongoClient` option if using `connect()` or `createConnection()`. See http://mongoosejs.com/docs/connections.html#use-mongo-client
Mongoose connected to the database
Constructing list of first names
Constructing list of last names
Constructing list of Addresses
(node:18802) DeprecationWarning: Mongoose: mpromise (mongoose's default promise library) is deprecated, plug in your own promise library instead: http://mongoosejs.com/docs/promises.html
🔥  Caught SIGINT, generating flamegraphbuffer.js:611
    return this.utf8Slice(0, this.length);
                ^

Error: "toString()" failed
    at Buffer.toString (buffer.js:611:17)
    at traceStacksToTicks (/home/jamie/.nvm/versions/node/v8.9.3/lib/node_modules/0x/lib/trace-stacks-to-ticks.js:11:6)
    at ChildProcess.<anonymous> (/home/jamie/.nvm/versions/node/v8.9.3/lib/node_modules/0x/platform/linux.js:104:18)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)
```

A bit of Googling suggests the problem may be a Buffer larger than the V8 string limit, see [this Node issue](https://github.com/nodejs/node/issues/3175).

davisjam:
The relevant source is:

```
function traceStacksToTicks (stacksPath) {
  const header = /(.+):(.+): ?$/
  var n = -1
  const stacks = fs.readFileSync(stacksPath)
    .toString()
    .split('\n')
    .reduce((stacks, line) => {
...
```

so I'm guessing the trouble is that the `stacksPath` file is too large.

If this sounds like the only problem here I will submit a PR to safely read the file line by line instead.
If there might be other troubles, let me know.

[...]

PR #121, opened 2018-04-17
Safely parse large trace stacks files

davisjam:
Problem:
If the file with trace stacks is larger than the maximum string length,
then traceStacksToTicks chokes on a call to toString.

Solution:
Split buffer on newlines via repeated indexOf rather than via
buf.toString.split('\n')

Test:
I ran assert.deepEqual to compare the stacks generated by the
old and new approaches. It said they were equivalent.

Fixes: #120

### DIFF
--- a/lib/trace-stacks-to-ticks.js
+++ b/lib/trace-stacks-to-ticks.js
@@ -4,12 +4,25 @@ const fs = require('fs')
 
 module.exports = traceStacksToTicks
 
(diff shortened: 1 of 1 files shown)


davidmarkclements:
amazing solution, thank you

you're right re visualization, I'm hoping to take a look at that soon 

I'll review this closely tomorrow and will merge asap

PR #131, opened 2018-07-04
Style fixes

goto-bus-stop:
- Fixes the height of the graph in FF
  For some reason `height: calc(100% - 66px)` does something totally
  different than `height: 100%`, but with `100vh` the behaviour is the
  same either way. This seems like a FF CSS bug maybe…
- Use floating + margins instead of absolute positions for top bar
  controls
  Default button and input sizes are different across browsers and OSes,
  by using floats + margins the positioning adapts to sizes instead of
  having things drawn on top of each other.

Before

image

After

image

### DIFF
--- a/visualizer/cmp/graph.js
+++ b/visualizer/cmp/graph.js
@@ -3,7 +3,7 @@
 module.exports = (render) => Object.assign(() => render`
   <chart 
     class='db overflow-y-scroll overflow-x-hidden relative' 
-    style='padding-left: 5%; padding-right: 5%; height: calc(100% - 66px)'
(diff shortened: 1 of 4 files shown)


davidmarkclements:
awesome thanks @goto-bus-stop – as a matter of process could you also add screenshots to show that the changes don't negatively affect chrome

pending screenshots, this LGTM

goto-bus-stop:
it's actually a small improvement on Chrome too! (on Arch Linux)

Before

image

After

image

davidmarkclements:
amazing, great work and thank you - merging

davidmarkclements:
v4.0.12 released

PR #135, opened 2018-07-20
Use timestamp instead of pid for isolate log for babel-node support

vizv:
babel-node spawn its own sub-process which generates a isolate log that
0x is not able to locate.

To resolve such a issue, timestamp might be used as a clue for locating
those isolate logs generated with non-standard 0x.

Tested with `0x -o -- babel-node --prof foo.js`.


vizv:
Not working when presets was added, close for now

vizv:
please ignore this PR, babel works with `0x -- node --require babel-register --require babel-polyfill index.js`

PR #137, opened 2018-07-24
fix regex on windows

mafintosh:
the regex stopped working for me on windows with node 10.6+


mafintosh:
Example windows filename for the record: `'isolate-000001A5B3359E40-4672-v8.json`

PR #143, opened 2018-08-08
Stretch height to fit tallest stack

kamil-mech:
Some stacks end up trimmed because they are taller than the viewport (can be confirmed by zooming out with `-` button)

## Screenshots of scroll-top

Before:
screen shot 2018-08-08 at 15 30 20

After:
screen shot 2018-08-08 at 15 30 44

## Screenshots of full scrollable content

Before:
screencapture-file-users-km-work-bubble-setup-node-clinic-examples-87001-clinic-flame-html-2018-08-08-15_28_43

After:
screencapture-file-users-km-work-bubble-setup-node-clinic-examples-87001-clinic-flame-html-2018-08-08-15_28_16

### DIFF
--- a/visualizer/index.js
+++ b/visualizer/index.js
@@ -16,11 +16,15 @@ module.exports = function (trees, opts) {
   const chart = graph()
   const tree = trees.unmerged // default view
   const categorizer = !kernelTracing && graph.v8cats
+  const cellCount = tree.children.map(getDeepest).reduce((prev, next) => Math.max(prev, next), 0)
+  const headerBarOffset = 55
+  const cellHeight = 18.5
   const flamegraph = fg({
     categorizer, 
     tree, 
     exclude: Array.from(exclude), 
-    element: chart
+    element: chart,
+    height: (cellCount * cellHeight) + headerBarOffset
   })
   const { colors } = flamegraph
 
@@ -39,4 +43,11 @@ module.exports = function (trees, opts) {
 
   document.body.appendChild(chart)
   document.body.appendChild(iface)
+}
+
+function getDeepest (node) {
+  if (!node.children) {
+    return 1
+  }
+  return node.children.map(getDeepest).reduce((prev, next) => Math.max(prev, next), 0) + 1
 }


davidmarkclements:
Does this increase scroll height?

Does it dynamically respond to graph height or are we just making it bigger, where an even higher graph will still be cropped?

kamil-mech:
> Does this increase scroll height?

Yes, this is reflected in the second screenshot on the right

> Does it dynamically respond to graph height or are we just making it bigger, where an even higher graph will still be cropped?

I tested multiple samples with taller and narrower stacks and did not find a single occurrence where it would overflow. Also it seems to cap at 256 rows

---

Meanwhile I have discovered the real root cause of the issue is in [maxDepth@d3-fg](https://github.com/davidmarkclements/d3-fg/blob/master/index.js#L777-L783)

It always returns `1` so we probably should fix it there + maybe add top offset option

Btw this PR should use `18` cell height as that would be in line with `d3-fg`

davidmarkclements:
There .5 was a hack, can't remember the reason 

What I meant was - does it increase the default scroll height 

Not finding an occurrence is not the same as there not being the possibility of an occurence - however if we can verify it caps at 256 rows (and if that's acceptable) then the max height should be 256*18

kamil-mech:
The `.5` was actually my mistake :)

> What I meant was - does it increase the default scroll height

It will increase (exceed 950), but only when there is a stack with at least ~50 rows
(50 * 18 = 900 in cells + 55 in header)
https://github.com/davidmarkclements/d3-fg/blob/master/index.js#L43-L45
\* Currently all graphs seem to cap height at 950

> Not finding an occurrence is not the same as there not being the possibility of an occurence

True. I have created code which should be able to produce 1K+ stack, but could not try it due to 256 cap

---

I suggest I open PR in `d3-fg`. It seems that [d3.tree](https://github.com/davidmarkclements/d3-fg/blob/master/index.js#L779) that we use [no longer takes in an iterator fn](https://github.com/d3/d3-hierarchy/blob/master/src/tree.js#L106), so we have to either:
1\) Port [getDeepest@thisPR](https://github.com/davidmarkclements/0x/blob/0c4054a253984f6ef2cbbe8f08b55c8561ae8365/visualizer/index.js#L48-L53) to replace [maxDepth@d3-fg](https://github.com/davidmarkclements/d3-fg/blob/master/index.js#L777-L783)
2) Find another `d3` abstraction that supports an iterator (any suggestions?)

Alternatively I can give this PR more tweaking to make sure it's consistent across all samples

kamil-mech:
FYI I have attempted to find where exactly it caps 256, but no luck yet

kamil-mech:
Once we merge https://github.com/davidmarkclements/d3-fg/pull/5 (or resolve in another way) then we can close this PR and open one that adds some `topOffset` (I think 55px will be ok) to prevent top bar from hiding top of the stack (tried with css first, but that resulted in some interesting side effects)

davidmarkclements:
closing in favour of https://github.com/davidmarkclements/d3-fg/pull/5

PR #146, opened 2018-08-11
Ignore visual studio code files

BridgeAR:


.npmignore
```diff
@@ -9,4 +9,5 @@ win
 0x-logo.png
 *.gif
 *.0x
-todo
\ No newline at end of file
+todo
+.vscode
```
kachkaev: Can the file end with `\n` given the opportunity? This will make future diffs cleaner.

PR #149, opened 2018-08-11
Fix error on resize

shuhei:
When I resize the browser window in `flamegraph.html`, I get the following error in console.

```
Uncaught TypeError: Cannot read property 'setAttribute' of null
    at window.addEventListener.debounce (flamegraph.html:11197)
    at later (flamegraph.html:9405)
```

The error happens because the code tries to resize an `<svg>` element while `d3-fg` doesn't use `<svg>` anymore.

This PR removes the line causing the error. Even without the line, the line above resizes the chart properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "ajv": "^6.0.1",
     "browserify": "^13.0.0",
     "concat-stream": "^1.5.2",
-    "d3-fg": "^6.1.0",
+    "d3-fg": "^6.2.3",
     "debounce": "^1.1.0",
     "debug": "^2.2.0",
     "end-of-stream": "^1.1.0",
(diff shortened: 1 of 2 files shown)


AlanSl:
Works as described, thanks! 

One little thing though which should probably be done at the same time - can you update the minimum version of `d3-fg` in `package.json`? Currently it's `^6.1.0`. `6.1.0` uses SVG, switching to Canvas happened in `6.2.0`. 

0x doesn't break if d3-fg 6.1.0 is used, but resize becomes sluggish. If we're changing the code assuming Canvas not SVG, we should specify latest 6.2, `^6.2.3`

shuhei:
Sure! Added a commit to update `d3-fg`.

PR #150, opened 2018-08-15
Fix resizing canvas after window resize

goto-bus-stop:
This was still trying to resize the svg instead of the canvas.

The setAttribute and update() lines need to be flipped because setting
the canvas width clears the contents; need a redraw afterward.


goto-bus-stop:
Dupe :laughing:

PR #151, opened 2018-08-15
Direct stack frame #-links + history

goto-bus-stop:
A flame graph made with this version is viewable at https://upload.clinicjs.org/public/509680025b4cc97cec5a7e2eaacdb90c10c5585ba70b7bb8656bc3985db468c1/8353.clinic-flame.html (see below for direct stack frame link examples)

This patch stores the current focused node in the URL hash, so you can
link to a specific stack. It does that by assigning each node an ID on
load. It's a counter that does a preorder traversal, this is a safe way
to do it because the data that is used is static, stack frame 352 will
be the same regardless of who opens the file.

It also implements the back/forward buttons using the `pushState()` API.
Only the merged/unmerged setting, the type visibility settings (c++, v8, etc)
and the focused node ID are stored, because those must be known in order
to show the correct node. Other stuff like search / tiers / whatnot is not
stored, I think that's less important anyway, if necessary people can just
use a direct stack frame link and tell the recipient that they should press
some buttons to get the perfect view. A tiny stack frame is hard to find but
a big button is not :)

The hash is read on load, elsewhere the history API is
used—`pushState()` and `onpopstate` are a little easier to deal with
than `hash` and `onhashchange`. (onpopstate is _only_ triggered by user
action for example so we don't have to check if we just pushState()d.)

Requires https://github.com/davidmarkclements/d3-fg/pull/7

## Examples

Direct link to a random stack frame I picked: https://upload.clinicjs.org/public/509680025b4cc97cec5a7e2eaacdb90c10c5585ba70b7bb8656bc3985db468c1/8353.clinic-flame.html#unmerged-2368-regexp+init+v8+native+cpp

Direct link to a C++ stack frame: https://upload.clinicjs.org/public/509680025b4cc97cec5a7e2eaacdb90c10c5585ba70b7bb8656bc3985db468c1/8353.clinic-flame.html#unmerged-2202-regexp+init+v8+native (would be hidden on load by default)

Direct link to a stack frame in the merged view: https://upload.clinicjs.org/public/509680025b4cc97cec5a7e2eaacdb90c10c5585ba70b7bb8656bc3985db468c1/8353.clinic-flame.html#merged-727-regexp+init+v8+native+cpp (default view is unmerged)

### DIFF
--- a/visualizer/actions.js
+++ b/visualizer/actions.js
@@ -110,24 +110,48 @@ function createActions ({flamegraph, state}, emit) {
   function pushState () {
     return (d) => {
       const { merged } = state.control
+      const excludeTypes = Array.from(state.typeFilters.exclude)
       const { nodesToIds } = merged ? mergedTags : unmergedTags
       const historyState = {
         merged,
         nodeId: nodesToIds.get(d),
+        excludeTypes
       }
       window.history.pushState(historyState, '', `#${stringifyHistoryState(historyState)}`)
     }
   }
 
+  // Jump to a state based on a history entry.
   function jumpToState () {
-    return ({merged, nodeId}) => {
+    return ({merged, nodeId, excludeTypes}) => {
       state.control.merged = merged
       state.typeFilters.enableInlinable = !merged
       state.key.enableOptUnopt = !merged
-      if (merged) flamegraph.renderTree(state.trees.merged)
-      else flamegraph.renderTree(state.trees.unmerged)
+
+      // Update type exclude state
+      // note that a user changing type exclusion does not add a new history entry,
+      // this is just necessary to make sure the correct types show up when linking
+      // to eg. a v8-internal stack frame.
+      const oldExclude = state.typeFilters.exclude
+      oldExclude.forEach((name) => {
+        if (!excludeTypes.includes(name)) { // it's still an array here 🙈
+          flamegraph.typeShow(name)
+        }
+      })
+      excludeTypes.forEach((name) => {
+        if (!oldExclude.has(name)) {
+          flamegraph.typeHide(name)
+        }
+      })
+      state.typeFilters.exclude = new Set(excludeTypes)
+      state.typeFilters.bgs = state.control.tiers
+        ? highlightTypeFilters()
+        : state.typeFilters.unhighlighted
+
+      flamegraph.renderTree(merged ? state.trees.merged : state.trees.unmerged)
       const { idsToNodes } = merged ? mergedTags : unmergedTags
       flamegraph.zoom(idsToNodes.get(nodeId))
+
       emit(state)
     }
   }
@@ -157,6 +181,6 @@ function tagNodesWithIds (data) {
   }
 }
 
-function stringifyHistoryState ({ merged, nodeId }) {
-  return `${merged ? 'merged' : 'unmerged'}-${nodeId}`
+function stringifyHistoryState ({ merged, nodeId, excludeTypes }) {
+  return `${merged ? 'merged' : 'unmerged'}-${nodeId}-${excludeTypes.join('+')}`
 }
--- a/visualizer/actions.js
+++ b/visualizer/actions.js
@@ -5,8 +5,11 @@ module.exports = createActions
 function createActions ({flamegraph, state}, emit) {
   state.typeFilters.bgs = state.typeFilters.unhighlighted
 
+  const unmergedTags = tagNodesWithIds(state.trees.unmerged)
+  const mergedTags = tagNodesWithIds(state.trees.merged)
+
   return {
-    search, control, zoom, typeFilters
+    search, control, zoom, typeFilters, pushState, jumpToState
   }
 
   function search () {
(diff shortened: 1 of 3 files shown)


AlanSl:
Nice! The approach to IDs seems to be robust and consistent from a little manual testing.

---

One observation on experimenting with this demo - the view filter states (deps, cpp, init etc) are stored in the hash, but only on selecting a frame. Toggling the filter buttons updates which frames are visible, but doesn't update the hash. 

This could give unexpected behaviour if user 1 clicks on a frame then toggles a filter then shares a hash: user 2 would see the frames that were visible before user 1 toggled the filter. Whereas if user 1 reached the exact same view by pressing the same buttons in a different order (toggle the filter _then_ click on the frame), user 2 does see the same frames as user 1.

I'd say it's probably worth also updating the hash on toggling the appropriate filter buttons, so that when user 1 sees a set of frames and shares the hash with user 2, user 2 always sees the same frames as user 1. 

---

Long term, I'm not sure about having some button states stored in the hash and some not (especially since there's currently no visual distinction between the types of buttons), but I think it's fair to keep this PR focused on frame visibility; then consider storing the tiers / optimized / unoptimized states as a separate feature.

goto-bus-stop:
Now updating the hash on type exclusion changes. I think I'll change the hash to a JSON string so it'll be easier to add things to it in the future, without having this custom parse/serialize thing. Initially I was hoping to keep the hash nice and simple but it's pretty messy with the type exclusions in there, so that basically went out of the window already.

AlanSl:
Nice. Here's an example sample generated after npm-linking the d3-fg PR mentioned above - https://upload.clinicjs.org/public/71fe6faf4b0dfeb059e10ca03eeac9c08e36c16eb583f893cc95756e44e35671/36543.clinic-flame.html

I think this approach is really good. Our target audience is devs and the JSON in the hash is readable and makes it clear what is going on. If it got really out of hand in future, we could consider something like encoding all the boolean states in an integer (`if (someInt >= 128) { someState = true; someInt =- 128 }; if (someInt >= 64) { anotherState = true; someInt =- 64 }` etc down to 1, or something fancy doing the same using bitwise) - but right now I think the verbose, human-readable hash is a bonus because users can clearly see what the feature offers, and what users they share the link with will get.

I was just about to ask how this handles URL escapes, but I see you have a commit for that already :-)

AlanSl:
I hadn't tried using the back button before - I just tried it on the demo above and it erased the hash, resulting in all of the toggle buttons being switched on.

Is this missing something in a dependency maybe (e.g. something new in d3-fg) or does updating the hash on back button use need to be added here?

goto-bus-stop:
The back button works correctly for me when moving between different frames, but when moving back to the initial state, it enables all buttons as you say. Is that the behaviour you're seeing too? It definitely is a bug.

visualizer/actions.js
```diff
@@ -100,7 +114,85 @@ function createActions ({flamegraph, state}, emit) {
...
   }
+
+  function pushState () {
+    return () => {
+      const { merged } = state.control
+      const excludeTypes = Array.from(state.typeFilters.exclude)
+      const { nodesToIds } = merged ? mergedTags : unmergedTags
+      const historyState = {
+        merged,
+        nodeId: nodesToIds.get(state.focusedNodeId),
```
AlanSl: There appears to be some inconsistency about `state.focusedNodeId`'s type: 

- This code seems to assume it's a node object (`tagNodesWithIds` sets `nodesToIds` keys as tree layers), and it appears that `flamegraph.on('zoom', (d) => { ... focusNode(d) }` in index.js:29-36 sets it as a node object 
 - But the hash JSON parser and `window.addEventListener('popstate'` listener in index.js set it as a numeric ID, and the property name suggests it will be a number

The code seems to work fine despite this - it seems that the `pushState` route consistently expects and gets `state.focusedNodeId` as a node object and the `jumpToState` route consistently expects and gets it as a number, and the two don't currently collide - but it seems like a pit trap to have a property that appears to be a numeric id, when by some routes it could actually be a node object?

AlanSl: Actually I think it is possible for the two routes to collide. To replicate:

- Click on some node. `state.focusedNodeId` is set as an object via the `pushState` route
- Toggle a visibility setting. Keeps the ID
- Undo the visibility setting with the back button. `state.focusedNodeId` is set as a number via the `jumpToState ` route
- Toggle another visibility setting. This unexpectedly erases the node Id from the hash - presumably, because it's going down the `pushState` route and doing `nodesToIds.get()` on a numeric id, which returns undefined?

visualizer/index.js
```diff
@@ -24,6 +24,25 @@ module.exports = function (trees, opts) {
...
+
+    focusNode(d)
+  })
+  window.addEventListener('popstate', (event) => {
+    userZoom = false
+    jumpToState(event.state || {
+      // No hash anymore, jump to root node (0) but don't change settings
+      merged: state.control.merged,
+      excludeTypes: Array.from(state.typeFilters.exclude),
+      nodeId: 0,
```
AlanSl: Tiny style nit: I personally don't have any issue with trailing commas in objects, but since there's currently no lint or style guide fixed for this repo, it's probably best to stick with the existing convention which appears to have been not adding trailing commas.

visualizer/actions.js
goto-bus-stop: good find, in my head `focusedNodeId` was always a number but clearly that's not what I actually wrote down :sweat_smile: thanks!

goto-bus-stop:
Updated, thanks for review :ok_hand:

AlanSl:
Great, changes make sense and newly created profile works as expected. LGTM.

PR #153, opened 2018-08-22
Accept FUNCTION SOURCE identifiers mid-line

### LINKED ISSUE #122: Profiled application prints output
When running with 0x, the Node-DC-EIS benchmark prints stuff.
With vanilla Node it does not.

```
Node-DC-EIS/Node-DC-EIS-cluster $ 0x server-cluster.js 
🔥  ProfilingStarting a single instance of Node.js process with (pid: 3230)
********************************
Server running at port: 9000
********************************
(node:3230) DeprecationWarning: `open()` is deprecated in mongoose >= 4.11.0, use `openUri()` instead, or set the `useMongoClient` option if using `connect()` or `createConnection()`. See http://mongoosejs.com/docs/connections.html#use-mongo-client
Mongoose connected to the database
Constructing list of first names
Constructing list of last names
Constructing list of Addresses
(node:3230) DeprecationWarning: Mongoose: mpromise (mongoose's default promise library) is deprecated, plug in your own promise library instead: http://mongoosejs.com/docs/promises.html
() {
            var args_ = Array.prototype.slice.call(arguments)
              , ret, total_, current_, next_, done_, postArgs;

            if (_current === _total) {
              
              next_ = function () {
                if (arguments[0] instanceof Error) {
                  return handleError(arguments[0]);
                }
                var args_ = Array.prototype.slice.call(arguments, 1)
                  , currPost
                  , postArgs;
                if (args_.length) hookArgs = args_;
                if (++current_ < total_) {
                  currPost = posts[current_]
                  if (currPost.length < 1)
                    throw new Error("Your post must have a next argument -- e.g., function (next, ...)");
                  postArgs = [once(next_)].concat(hookArgs);
                  return currPost.apply(self, postArgs);
                } else if (typeof lastArg === 'function'){
                  // All post handlers are done, call original callback function
                  return lastArg.apply(self, arguments);
                }
              };

              // We are assuming that if the last argument provided to the wrapped function is a function, it was expecting
              // a callback.  We trap that callback and wait to call it until all post handlers have finished.
              if(typeof lastArg === 'function'){
                args_[args_.length - 1] = once(next_);
              }

              total_ = posts.length;
              current_ = -1;
              ret = fn.apply(self, args_); // Execute wrapped function, post handlers come afterward

              if (total_ && typeof lastArg !== 'function') return next_();  // no callback provided, execute next_() manually
              return ret;
            }
          }
--- END ---
() {
    var pDomain = promise.emitter.domain;
    if (pDomain && pDomain !== process.domain) pDomain.enter();
    try {
      var x = fn.apply(undefined, boundHandler.args);
    } catch (err) {
      promise.reject(err);
      return;
    }
    resolve(promise, x);
  }
--- END ---
(key) {
    result[key] = Object.getOwnPropertyDescriptor(object, key);
    // Assume these are schema paths, ignore them re: #5470
    if (result[key].get) {
      delete result[key];
      return;
    }
    result[key].enumerable = ['isNew', '$__', 'errors', '_doc'].indexOf(key) === -1;
  }
--- END ---
🔥  Process exited, generating flamegraph
```

Why is this happening?

goto-bus-stop:
Reading v8 output from the stdout stream of a child process is sometimes
unreliable, chunks may be missing. The next `FUNCTION SOURCE` can be in
the middle of a previous `FUNCTION SOURCE`, cutting it off. The impact
can be alleviated slightly by accepting `FUNCTION SOURCE` identifiers
that occur in the middle of a line. It's not 100% accurate, because the
missing chunks may contain other `FUNCTION SOURCE` blocks, so we can
still be missing out on stuff. But, at least we don't dump function
source code to the shell anymore.

When this happens, the previous FUNCTION SOURCE block will not have been
closed by an END line, but END lines only set
`stdoutIsPrintOptSourceOutput` to false, and FUNCTION SOURCE lines set
it to true, so we don't have to do anything there.

I tightened the FUNCTION SOURCE regex a bit for this, so that a literal
`--- FUNCTION SOURCE` occuring inside a function's source code (eg a
string or regex) won't break. It's still possible to trick 0x by doing
something like:
```
// --- FUNCTION SOURCE (...) id{...} start{...}
```
but I hope nobody does that… :sweat_smile: 

Related to but does not quite fix #122

### DIFF
--- a/platform/v8.js
+++ b/platform/v8.js
@@ -123,7 +123,16 @@ function collectInliningInfo (sp) {
         inlined[key].push(lastOptimizedFrame)
         cb()
         return
-      } else if (/^--- FUNCTION SOURCE /.test(s)) {
+        // Reading v8 output from the stdout stream is sometimes unreliable. The next
+        // FUNCTION SOURCE can be in the middle of a previous FUNCTION SOURCE, cutting
+        // it off. The impact can be alleviated slightly by accepting FUNCTION SOURCE
+        // identifiers that occur in the middle of a line.
+        // The previous FUNCTION SOURCE block will not have been closed, but END lines
+        // only set `stdoutIsPrintOptSourceOutput` to false, so we don't have to do
+        // anything here. If the END logic changes the below may need to change as well.
+        //
+        // ref: https://github.com/davidmarkclements/0x/issues/122
+      } else if (/--- FUNCTION SOURCE \(.*?\) id\{\d+,-?\d+\} start\{\d+\} ---\n$/.test(s)) {
         stdoutIsPrintOptSourceOutput = true
         const [match, file, fn = '(anonymous)', id, ix, pos] = /\((.+):(.+)?\) id\{(\d+),(-?\d+)\} start\{(\d+)}/.exec(s) || [false]
         if (match === false) return cb()


mcollina:
I think we can claim this fixed #122. It's the best we can do, and I don't think the V8 team would be very keen in fixing this bug (it's on them).

PR #154, opened 2018-08-24
Send SIGTERM to subprocess on second Ctrl+C

### LINKED ISSUE #136: 0x may exit without generating flamegraph on SIGINT
With node v8.11.3 installed with nvm in Gentoo recently, 0x (4.1.0) occasionally exit without generating flamegraph on SIGINT (generated with Ctrl + C).

I did some investigation, and I think the issue is related to this line:

https://github.com/davidmarkclements/0x/blob/cbd5c309d0adea0d2cec8d7bf8ccb967670ebd4e/platform/v8.js#L49

I tried to put the following debug print just before the `Promise race`:

```javascript
process.on('SIGINT', () => console.log('|SIGINT|'))
```

and when I press Ctrl + C it prints two `|SIGINT|` however it won't exit without generating flamegraph anymore.

I have no idea why it emits two `SIGINT` events to `process` and how to fix this issue (or is it just me)?

goto-bus-stop:
This fixes an issue where, if the subprocess was stuck on a sync operation, pressing Ctrl+C would generate a flame graph and exit 0x, but the subprocess would keep spinning on in the background.

I used:

```js
// loop.js
while (true);
```
```bash
./cmd.js loop.js
```

Previously, doing Ctrl+C would generate the flame graph, and then 0x would keep running afterward because the subprocess hadn't ended. Sending another Ctrl+C would terminate 0x, as it no longer has a SIGINT handler, but the subprocess would keep running.

With this patch, we catch two SIGINTs—the first uses the `stdio[5].destroy()` soft-close mechanism like before, and the second sends a SIGTERM to the subprocess. 0x suggests sending a second Ctrl+C if soft-closing takes a very long time.
As a last-resort measure, a SIGTERM is also sent to the subprocess when 0x itself exits. (perhaps this should just do `proc.kill()` to be absolutely sure?) This can happen when using --on-port because the flamegraph starts being generated as soon as the --on-port process ends, rather than when the process being monitored ends. (hmm, perhaps it should wait for the monitored process to end too?)

<strike>This branch is a bit of a mess because I was trying and failing to address #136. I'll squash it in a bit.</strike>done. Note this patch does not solve #136.

### DIFF
--- a/examples/rest-api/package.json
+++ b/examples/rest-api/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "wrk -d 120 -c 100 http://localhost:3000/a & wrk -d 120 -c 100 http://localhost:3000/b & wrk -d 120 -c 100 http://localhost:3000/c"
+    "test": "wrk -d 5 -c 100 http://localhost:3000/a & wrk -d 5 -c 100 http://localhost:3000/b & wrk -d 5 -c 100 http://localhost:3000/c"
   },
   "author": "David Mark Clements",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "stress-rest-example": "cd examples/rest-api && npm test",
+    "start": "./cmd.js --on-port 'npm run stress-rest-example' -- node examples/rest-api",
     "lint": "standard"
   },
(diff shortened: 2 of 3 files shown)


mcollina:
Nice job.

mcollina:
> hmm, perhaps it should wait for the monitored process to end too?

I think it might be a good idea.

goto-bus-stop:
Yeah, when I wrote that down I immediately went to work on that because it seemed better :smile: just pushed that change!

PR #159, opened 2018-08-25
Minify the JS in the generated file

BridgeAR:


visualizer/html.js
```diff
@@ -13,6 +14,11 @@ function html (opts) {
   const { htmlPath, script } = opts
   const stdout = (opts.htmlPath === '-' || !opts.htmlPath)
 
+  const res = minify(script, { mangle: false })
+  if (res.error) {
+    throw res.error
+  }
```
mcollina: This should be behind an option and a CLI flag.

BridgeAR:
Closing as this will actually not improve much (the overall size will only drop a tiny bit).

Issue #119, opened 2018-04-12
Testing v8 version different from logging version


BridgeAR:
When trying to use Node.js master I always get the an error without the possibility of getting a flamegraph.

```
0x -o -- ./out/Release/node test.js
// ...
// 🚫  Testing v8 version different from logging version
```

Trying to circumvent the issue by first only collecting the information and then to generate the flamegraph from that results in a different error:

```
0x --collect-only -- ./out/Release/node t.js
0x --visualize-only ./folder.0x
// 🔥  Profiling
// 🚫  Target subprocess error, code: 9
```

davidmarkclements:
this is a problem with Node master

the V8 version reported by `process.versions.v8` and the v8 version in the log file header 
are different

you can work around it by changing the log file header to the version reported by `process.versions.v8`

came across this problem a while, checked again and it had been fixed in Node master, and now it's seems to be back again

BridgeAR:
I am not sure in what way this is broken in Node core. I guess you mean e.g., `'6.6.346.24-node.5'` instead of `'6.6.346.24'`?

That is not a bug but a feature. We add extra patches to Node.js on top of the regular V8 version and should not change the original version at all. That is only done in case we pull in a new V8 version. I guess that is also what happened in the moment you checked and found that it was `fixed`.

paambaati:
Is this possible? I'm trying to profile worker threads' performance on the latest master (`v11.0.0-pre`), and `0x` won't generate flamegraphs.

BridgeAR:
Ping @davidmarkclements. I am running into this again as I would like to use 0x for Node core master.

PR #168, opened 2018-09-17
Make sure the same Node.js executable is used

BridgeAR:
So far the child process spawned to vizualize the data could be a
different one than the one used to run 0x. Now it's always the same
and there should be no mismatch left.

Fixes: #119

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -13,7 +13,7 @@ const { version } = require('./package.json')
 
 const defaultBanner = `
   0x ${version}
-  
+
   0x [flags] -- node [nodeFlags] script.js [scriptFlags]
 
 `
@@ -24,7 +24,9 @@ if (module.parent === null) {
     debug(err)
     process.exit(err.code || 1)
   })
-} else module.exports = cmd
+} else {
+  module.exports = cmd
+}
 
 async function cmd (argv, banner = defaultBanner) {
   var args = minimist(argv, {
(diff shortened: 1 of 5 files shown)


cmd.js
```diff
@@ -97,7 +99,7 @@ function parseSubprocessCommand (args) {
       0x [flags] -- node [nodeFlags] script.js [scriptFlags]
     `)
   }
-  var pathToNodeBinary = false
+  var pathToNodeBinary = process.argv[0]
```
mcollina: I think this should *not* be `process.argv[0]`, but rather `./path/to/node` in `0x -- ./path/to/node script.js`.

lib/v8-log-to-ticks.js
```diff
@@ -9,9 +9,9 @@ const pump = require('pump')
 
 module.exports = v8LogToTicks
 
-function v8LogToTicks (isolateLogPath) {
+function v8LogToTicks (isolateLogPath, node) {
   const isJson = extname(isolateLogPath) === '.json'
-  const sp = isJson || spawn(process.argv[0], [
+  const sp = isJson || spawn(node, [
```
mcollina: This should be the one passed, not the one 0x is using itself. The reason is that we are somewhat confident that a Node.js release could process its own data, but not across multiple release lines.

cmd.js
BridgeAR: This is the fallback in case there is no extra path specified. It will then be used throughout the application.

But after looking at it again, I found that indeed that there was one case that was not covered: using it like `path/to/node cmd.js -- node script.js`. I just pushed that as well with a few other minor improvements.

lib/v8-log-to-ticks.js
BridgeAR: That is exactly what my change does.

PR #171, opened 2018-09-19
Fix writing meta.json in --collect-only mode

goto-bus-stop:
This was causing the failures @alansl was seeing in the flame tests.

Flame uses collectOnly and then visualizeOnly to create the flame graph,
rather than calling 0x once and letting it do the entire thing. After
9a1d4f069b629e256ef10b9fbb031583e66cda3d, meta.json was only written if
neither collectOnly nor visualizeOnly were used, which means it'd always
be missing with Flame.


BridgeAR:
Uh, makes sense. Sorry about that.

BridgeAR:
We definitely need more tests in place.

PR #172, opened 2018-09-19
Fixed support for Node 8.12.0.

mcollina:


lib/v8-log-to-ticks.js
```diff
@@ -22,8 +43,11 @@ function v8LogToTicks (isolateLogPath, node) {
     // it's an error in the V8 internal tick profiler)
     // ignore them.
     const ignore = /unknown code state:/
+    // pump(sp.stderr, process.stderr)
```
BridgeAR: Obsolete comment? :)

lib/v8-log-to-ticks.js
```diff
@@ -3,11 +3,32 @@
...
 const fs = require('fs')
 const pump = require('pump')
 
-module.exports = v8LogToTicks
+const readFile = promisify(fs.readFile)
+const writeFile = promisify(fs.writeFile)
+
+module.exports = process
+
+function process (isolateLogPath, node) {
```
BridgeAR: Can we use a different name? Having the same name as global seems not ideal.

lib/v8-log-to-ticks.js
```diff
@@ -3,11 +3,32 @@
...
+  return filterLinesThatAreTooLong(isolateLogPath).then((filtered) => {
```
BridgeAR: I suggest to only execute this when it's necessary (by checking the Node.js version).

lib/v8-log-to-ticks.js
```diff
@@ -3,11 +3,32 @@
...
+
+// long lines make the --preprocess crash. We are just filtering them beforehand
+// as Node.js 8.12.0 would crash if there is any error in the preprocess script
+function filterLinesThatAreTooLong (isolateLogPath) {
+  const dest = isolateLogPath + '.filtered'
+  // TODO use streams
+  return readFile(isolateLogPath, 'utf8').then((data) => {
+    data = data.split('\n').filter((s) => s.length < 1024).join('\n')
+    return writeFile(dest, data)
+  }).then(() => dest)
```
BridgeAR: Suggestion: use async / await.

mcollina:
PTAL

mcollina:
Those lines cause the processor to fail, and we silence that error. I think this is just avoid future problems.

zbjornson:
@mcollina this doesn't seem to always fix the problem -- I'm still getting the error with lines > 1024 characters removed. Maybe something else besides long lines causes `printErr` to be invoked?

mcollina:
Can you upload an example to reproduce? Thanks!

Il giorno mer 17 ott 2018 alle 04:00 Zach Bjornson <notifications@github.com>
ha scritto:

PR #173, opened 2018-09-19
Do not require pathToNodeBinary to be set

BridgeAR:
This was the case when 0x was used in a programatic way.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,6 @@ const { sun, linux, windows, v8 } = require('./platform')
 const debug = require('debug')('0x')
 const { join, isAbsolute, relative } = require('path')
 const fs = require('fs')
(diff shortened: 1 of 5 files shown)


AlanSl:
Still getting the error when using this branch in Clinic Flame. Source of error is index.js:107 `await v8LogToTicks(src, pathToNodeBinary)`, we should apply fallback in the `visualize` function too.

Might be other places also.

AlanSl:
`...name, pathToNodeBinary = process.execPath })` in `visualize` function arguments seems to fix flame private master.

BridgeAR:
@AlanSl please check again. This should fix it.

AlanSl:
LGTM. Fixes both public and private and is much cleaner than my suggestion.

PR #177, opened 2018-09-21
Expose stuff for clinic

goto-bus-stop:
- Adds a `writeTicks` option that saves the processed stack sample information to a JSON file `ticks.json`.
- Exports the `ticksToTree` function so that a custom visualizer can turn the stack sample information into the tree 0x uses.

<strike>This may need changes depending on how Clinic Flame dev goes over the next week or so, but opening this PR now mostly for visibility</strike> Seems a good start!


index.js
```diff
@@ -4,6 +4,7 @@ const { sun, linux, windows, v8 } = require('./platform')
 const debug = require('debug')('0x')
 const { join, isAbsolute, relative } = require('path')
 const fs = require('fs')
+const { promisify } = require('util')
```
mcollina: why is this needed?

goto-bus-stop: oops, it's not. thanks!

docs/api.md
```diff
@@ -121,4 +121,27 @@ Default: false
...
+const ticks = JSON.parse(fs.readFileSync('{pid}.0x/ticks.json', 'utf8'))
+```
+
+This function returns an object with properties:
+ - `merged` - a tree of stack frames, where each node includes both the
+   optimized and unoptimized calls for that frame.
+ - `unmerged` - a tree of stack frames, where each node refers to either the
+   optimized or unoptimized calls.
+
+Each node has properties:
```
BridgeAR: We add some extra properties on top of the existing ones from time to time (e.g., `isInit`. I am not sure if we actually need that property and I believe we do not but I did not yet test that properly and I do not want to remove it without having proper tests in place).

I suggest to add a comment that there might be some further properties as well that are not there by default.

index.js
```diff
@@ -15,6 +15,8 @@ const platform = process.platform
 const { tidy, noop, isSudo } = require('./lib/util')
 
 module.exports = zeroEks
+// Expose for custom visualizations
+module.exports.ticksToTree = ticksToTree
```
mcollina: can you move this to a separate file, and maybe add a unit test (maybe using tap snapshots?)

goto-bus-stop:
Ow, need to do something about `ticksToTree`'s handling of inlined functions, too. This is not done yet :(

docs/api.md
```diff
@@ -133,6 +133,20 @@ custom visualizations. After a `collectOnly` run, the `ticks` data is stored in
 const ticks = JSON.parse(fs.readFileSync('{pid}.0x/ticks.json', 'utf8'))
 ```
 
+`opts` is an object with optional properties:
+
+ - `inlined` - data about inlined functions. 0x stores this in the `meta.json`
+   file, which you can read like:
+
+   ```js
+   const { inlined } = JSON.parse(fs.readFileSync('{pid}.0x/meta.json', 'utf8'))
```
BridgeAR: Suggestion: use the following instead:

```js
const { inlined } = require('{pid}.0x/meta.json')
```

docs/api.md
```diff
@@ -133,6 +133,20 @@ custom visualizations. After a `collectOnly` run, the `ticks` data is stored in
...
+   ```
+
+ - `mapFrames(frame)` - See [mapFrames](#mapframes-function).
+ - `pathToNodeBinary` - Path to the Node.js executable. Defaults to
```
BridgeAR: We might want to shorten the name if we expose this variable to e.g., `binaryPath`.

docs/api.md
goto-bus-stop: My preference is for the more verbose `fs`+`JSON` in this situation. in practice this will almost always be a dynamic file path. I'm not a big fan of recommending something like `require(path.join(pathTo0xData, 'meta.json'))`. To me, using fs and then JSON.parse looks like "I'm reading a data file" while `require` looks like "I'm importing a module".

Is there some best practice or benefit to using `require` that I'm not aware of? :smile:

docs/api.md
goto-bus-stop: I kinda agree but the main 0x function has an option by this same long name :(

docs/api.md
BridgeAR: It's just shorter and less to read. So it's more intuitive for me. I guess it depends on how well you know the internals of require.

docs/api.md
BridgeAR: Oh, true. I forgot about that. Too bad. We have to wait for a major to rename it.

docs/api.md
mcollina: I prefer `fs.readFileSync` as well. require caches things, while `fs.readFileSync` does not.

PR #178, opened 2018-09-22
Improve init

BridgeAR:
This improves the init detection in two ways: it should stop detecting some native frames as init and it should detect esm modules properly.

I was not able to reproduce the native frames but it should still work properly.

At the moment the esm detection adds the next tick file to the init phase as well (`internal/process/next_tick.js`) and I am somewhat on the edge of having it in INIT or not. Opinions?

### DIFF
--- a/lib/ticks-to-tree.js
+++ b/lib/ticks-to-tree.js
@@ -5,6 +5,8 @@ const { join } = require('path')
 const debug = require('debug')('0x: ticks-to-tree')
 
 const preloadDirRx = RegExp(join(__dirname, 'preload'))
+const internalModuleRegExp = /^.?(?:\(anonymous\)|internalBinding|NativeModule[^ ]*) [^/\\][a-zA-Z0-9_/\\-]+\.js:\d+:\d+$/
+const nodeBootstrapRegExp = / internal\/bootstrap.+\.js:\d+:\d+$/
 
 module.exports = ticksToTree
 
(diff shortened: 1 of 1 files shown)


lib/ticks-to-tree.js
```diff
@@ -25,7 +27,7 @@ function ticksToTree (ticks, mapFrames, inlined, binary) {
 
   // Spawn a file that throws an error to get the loading stack.
   // Then create a regular expression that matches all those files.
-  const child = spawnSync(binary, [join(__dirname, 'loading-stacks')])
+  const child = spawnSync(binary, ['--experimental-modules', join(__dirname, 'loading-stacks')])
```
mcollina: We should be working even on Node.js versions where this flag is not available, as this will error otherwise. How can we handle that?

BridgeAR: 0x support is from 8.5 upwards and that is exactly the version where support for esm is added as well. That's why I did not add a special casing.

It would be easy to just check for the version otherwise.

mcollina: Thanks! As for future proofing it, I would recommend to check if the process failed, and then rerun without the flag.

BridgeAR: Done.

BridgeAR: Done.

PR #185, opened 2018-10-03
feat: add source map support and relative paths

pimlie:
This PR adds support for displaying file names and line numbers from source maps instead of dist packages. This can help to create a better understanding of what is when happening.

As I was working on that, I also noticed the file names can become quite long (at least in my environment). Therefore I also added a relative-path option so absolute file names will be displayed relative to that path. 

I have added an example for the source-map.


cmd.js
```diff
@@ -85,6 +90,28 @@ async function cmd (argv, banner = defaultBanner) {
...
+    args.sourceMaps = null
+  } else {
+    if (args.sourceMaps !== '') {
+      const sourceMapFile = resolve('.', args.sourceMaps)
+      if (fs.existsSync(sourceMapFile)) {
+        try {
+          args.sourceMaps = JSON.parse(fs.readFileSync(sourceMapFile))
+        } catch (err) {
+          console.error(`Couldn't parse source map file ${sourceMapFile}, is it valid json?`)
+        }
```
BridgeAR: I suggest to remove the try / catch. Not only could it be invalid json, it could also be a permission error and a wrong path. On top of that the program should close when some of the arguments can not be properly identified.

cmd.js
```diff
@@ -85,6 +90,28 @@ async function cmd (argv, banner = defaultBanner) {
...
+      }
+    }
+    if (typeof args.sourceMaps !== 'object') {
+      args.sourceMaps = {}
+    }
```
BridgeAR: That is a bit to much magic for me. I suggest to either throw an error or to remove the `if` so a regular error would occur later on.

examples/source-maps/rollup.config.js
```diff
@@ -0,0 +1,35 @@
+const { terser } = require('rollup-plugin-terser')
+const { copyFileSync } = require('fs')
+
+const after = () => {
+  return {
+    generateBundle (outputOptions, bundle, isWrite) {
+      setTimeout(() => {
+        copyFileSync('./index1.min.js.map', './index2.min-map.js')
+      }, 1)
```
BridgeAR: What is this timeout for?

readme.md
```diff
@@ -239,6 +239,22 @@ Show output from DTrace or perf(1) tools.
...
+### --source-maps
+
+If a javascript library has a source map (a sourceMappingUrl comment on the last
+line), list the source file and position instead of the original file.
+
+A path to a json file can be supplied to define source maps which can not be
+resolved automatically. The JSON file should list an object with library name as
+key and the source map file as value.
+
+### --relative-path
```
BridgeAR: Is this really necessary? The source map support should also be possible without that, right?
AFAIK `output-dir` should already be enough in this case.

BridgeAR:
@pimlie this is great! Thanks a lot for doing this. I just had a glimpse at the code and will look at it in more depth later on.

cmd.js
pimlie: The magic here is with regards to validation with schema.json. The argument from the cli will always be a string, so we have to make sure its an object because otherwise ajv will complain :)

examples/source-maps/rollup.config.js
pimlie: Thats probably not required anymore, at first I also create a source map for index2 and then renamed that one to one that couldnt be resolved automatically. But this didnt work well because the generated files didnt exist yet in this hook, they did in nextTick (which I implement here with the timeout). I can check if I can remove this now

readme.md
pimlie: Not sure what you mean, but the idea was that you can both use `--source-maps` as `-source-maps <file>`. In the first case source maps are resolved either by the sourceMapUrl comment or because a file ending on `.map` exists (eg `dist.js.map` for `dist.js`).
In the latter you can specify which source map to use in case a library doesnt specify one. Not sure if this will ever be the case, but as the name uses `url` this could also potentially mean (I guess) a library points to a remote source map on a webserver. If that would be the case, then you could specify it manually.

cmd.js
pimlie: Will do

usage.txt
```diff
@@ -60,6 +60,19 @@
...
 
+  --source-maps <file>    If a javascript library has a source map (a sourceMappingUrl
+                          comment on the last line), list the source file and position
+                          instead of the original file.
+
+                          <file> can be used to define source maps which can not be
+                          resolved automatically. It should be a JSON file of an object
+                          with <library name> as key and <source map file> as value.
+
+  --relative-path <dir>   Make the listed file names relative to this dir.
```
davidmarkclements: this is regardless of whether source maps are used?

usage.txt
```diff
@@ -60,6 +60,19 @@
                           
                           Default: false
 
+  --source-maps <file>    If a javascript library has a source map (a sourceMappingUrl
```
davidmarkclements:
```suggestion
  --source-maps <file>    If a JavaScript library has a source map (a sourceMappingUrl
```

readme.md
```diff
@@ -239,6 +239,22 @@ Show output from DTrace or perf(1) tools.
 
 Default: false
 
+### --source-maps
+
+If a javascript library has a source map (a sourceMappingUrl comment on the last
```
davidmarkclements:
```suggestion
If a JavaScript library has a source map (a sourceMappingUrl comment on the last
```

readme.md
```diff
@@ -239,6 +239,22 @@ Show output from DTrace or perf(1) tools.
...
+line), list the source file and position instead of the original file.
+
+A path to a json file can be supplied to define source maps which can not be
```
davidmarkclements:
```suggestion
A path to a JSON file can be supplied to define source maps which can not be
```

lib/ticks-to-tree.js
```diff
@@ -10,19 +14,24 @@ const nodeBootstrapRegExp = / internal\/bootstrap.+\.js:\d+:\d+$/
...
-    value: 0,
-    top: 0,
-    children: []
-  }
-  const unmerged = {
-    name: 'all stacks',
-    value: 0,
-    top: 0,
-    children: []
+let globalOptions
```
davidmarkclements: can you please explain the rationale behind this?

if we need to maintain singleton state, do it at the top level 

options can be passed down through functions – can we do that instead?

lib/ticks-to-tree.js
```diff
@@ -89,75 +99,183 @@ function ticksToTree (ticks, options = {}) {
               if (callerMatch) S += 2
             }
           }
+          if (sourceMaps !== null) {
+            jsLibraries[key.split(' ')[1]] = true
```
davidmarkclements:
```suggestion
            const [, lib] = key.split(' ')
            jsLibraries[lib] = true
```

lib/ticks-to-tree.js
```diff
@@ -89,75 +99,183 @@ function ticksToTree (ticks, options = {}) {
...
+async function loadSourceMap (fileName, sourceMaps) {
+  if (typeof loadedSourceMaps[fileName] === 'undefined') {
+    const sourceMapFiles = [
+      sourceMaps[fileName],
+      sourceMaps[basename(fileName)],
+      fileName + '.map' // dist.js -> dist.js.map
+    ]
+
+    if (existsSync(fileName)) {
+      const lastLine = await readLastLines.read(fileName, 1)
```
davidmarkclements: is there ever the possibility of situation where the last line is blank and the second to last line is the sourceMapUrl? Or do all tools always ensure that the sourceMapUrl is the final line, with no blank line?

package.json
```diff
@@ -45,14 +45,18 @@
...
     "semver": "^5.5.1",
     "single-line-log": "^1.0.1",
+    "source-map": "^0.7.3",
     "split2": "^3.0.0",
     "tachyons": "^4.9.1",
     "through2": "^2.0.1",
     "which": "^1.2.4"
   },
   "devDependencies": {
+    "rollup": "^0.66.2",
```
davidmarkclements: can you please pull the rollup dev deps into a separate `package.json` in the `examples/source-maps` folder

lib/ticks-to-tree.js
pimlie: Yes, looking at it again I am not sure why I even used that globalOptions variable as it only needs to pass to the renameFrame method

lib/ticks-to-tree.js
pimlie: The [rfc](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?pli=1#) states: `The generated code may include a line at the end of the source, with the following form:`. I read this as sourceMappingURL _must be_ the last line, but its a small change to read the last 2 lines instead to accomodate for an accidental line-break.

pimlie:
@davidmarkclements Sorry for the delay in working on this, but I finally had some time to implement your requested changes. Let me know if you have any remarks! :champagne:

pimlie:
As this issue has become quite stale I am closing it

Issue #186, opened 2018-10-08
Do not classify net _tickCallback as INIT


mcollina:
executing `process.nextTick` is currently misrecognized as INIT in all cases. However, we should not recognize it as INIT after boot has completed.

cc @BridgeAR

BridgeAR:
I mentioned that in the PR that added esm support: https://github.com/davidmarkclements/0x/pull/178#issue-217449654. There is no specific boot phase that is detected. It is just all parts that have something to do with booting up. I'll filter the next ticks now.

mcollina:
Thanks!

Il giorno lun 8 ott 2018 alle ore 15:30 Ruben Bridgewater <
notifications@github.com> ha scritto:

PR #187, opened 2018-10-08
Filter next tick

BridgeAR:
Before:
image

After:
image

This also moves the main filter logic into the executed function.
This seemed cleaner instead of having to do this in the receiving
part.

Fixes: https://github.com/davidmarkclements/0x/issues/186

### DIFF
--- a/lib/ticks-to-tree.js
+++ b/lib/ticks-to-tree.js
@@ -3,6 +3,7 @@
 const { spawnSync } = require('child_process')
 const { join } = require('path')
 const debug = require('debug')('0x: ticks-to-tree')
+const assert = require('assert')
 
 const preloadDirRx = RegExp(join(__dirname, 'preload'))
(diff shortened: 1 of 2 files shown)


mcollina:
Thanks!!

lib/ticks-to-tree.js
```diff
@@ -28,23 +29,16 @@ function ticksToTree (ticks, options = {}) {
...
-  let regExpStr = '('
-  let separator = ''
-  for (const frame of new Set(stacks)) {
-    regExpStr += `${separator} ${frame.replace(/\\/g, '\\\\').replace(/\./g, '\\.')}\\d+:\\d+`
-    separator = '|'
-  }
-  regExpStr += ')$'
-  const regExp = new RegExp(regExpStr)
+  const stackRegExpStr = stackChild.stdout.toString()
+  assert(stackRegExpStr)
```
goto-bus-stop: @BridgeAR This assertion fails for me (and CI) on Node 8 in the flame-private tests. Do you have an idea why that might be?

AlanSl: Relevant flame-private issue: https://github.com/nearform/node-clinic-flame-private/issues/19

Issue #190, opened 2018-10-18
Invalid regular expression & Syntax Error


Zayelion:
Windows 7
Running in VS code CMD terminal 
Node 8.5.0

```
Process exited, generating flamegraphbufferSyntaxError:
Invalid regular expression: / C:\d+:\d+| x86)\\Nodist\\bin\\node_modules\\0x\\lib\\random-require\.js:\d+:\d+| module\.js:\d+:\d+| internal/module\.js:\d+:\d+| x86)\\Nodist\\bin\\node_modules\\0x\\lib\\loading-stacks\.js:\d+:\d+| bootstrap_node\.js:\d+:\d+$
/: Unmatched ')'
    at RegExp (<anonymous>)
    at ticksToTree (C:\Program Files (x86)\Nodist\bin\node_modules\0x\lib\ticks-to-tree.js:41:18)
    at render (C:\Program Files (x86)\Nodist\bin\node_modules\0x\lib\render.js:19:17)
    at generateFlamegraph (C:\Program Files (x86)\Nodist\bin\node_modules\0x\index.js:95:22)
    at zeroEks (C:\Program Files (x86)\Nodist\bin\node_modules\0x\index.js:59:18)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
Terminate batch job (Y/N)?
```

mcollina:
Thanks for reporting!

mcollina:
I think Node 8.5.0 has a bug, everything is working as expected in Node 8.12. Would you mind updating?

mcollina:
Actually this is a bug and broken in latest 0x. I'll get a fix soon.

Zayelion:
Thank you for your quick response. With node 8.12 issue persist.

PR #191, opened 2018-10-19
Escape __dirname before generating regexp

mcollina:
Fixes https://github.com/davidmarkclements/0x/issues/190

### DIFF
--- a/lib/ticks-to-tree.js
+++ b/lib/ticks-to-tree.js
@@ -4,8 +4,9 @@ const { spawnSync } = require('child_process')
 const { join } = require('path')
 const debug = require('debug')('0x: ticks-to-tree')
 const assert = require('assert')
+const escape = require('escape-string-regexp')
 
-const preloadDirRx = RegExp(join(__dirname, 'preload'))
+const preloadDirRx = RegExp(join(escape(__dirname), 'preload'))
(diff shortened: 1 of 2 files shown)


lib/ticks-to-tree.js
```diff
@@ -4,8 +4,9 @@ const { spawnSync } = require('child_process')
 const { join } = require('path')
 const debug = require('debug')('0x: ticks-to-tree')
 const assert = require('assert')
+const escape = require('escape-string-regexp')
 
-const preloadDirRx = RegExp(join(__dirname, 'preload'))
+const preloadDirRx = RegExp(join(escape(__dirname), 'preload'))
```
goto-bus-stop: This won't escape `windows-path\preload`, but I don't think that's a problem… `escape(join())` would make sure the entire thing is escaped, regardless

goto-bus-stop: Actually escaping before joining does introduce more backslashes which _may_ be a problem for unicodey windows paths? I'm not really familiar with them.

mcollina: escaping after joining does *not* work. I tried that first.

goto-bus-stop: Ah, that's unintuitive. Alright :+1:

Issue #192, opened 2018-10-22
Name strings on Windows have many characters missing (not just characters following `\`)


AlanSl:
Problem found in 0x 4.5.2, 4.5.1, 4.5.0 and 4.4.0 on Windows 10; and @mafintosh has replicated and found the same in 4.5.2 and 4.5.0 on Windows Server 2016. 

For example: 

- For a `cpp` frame:
  - expected: `C:\Program Files\nodejs\node.exe` 
  - shows as: `C:am Files odejs ode.exe`
- For an `app` frame:
  - expected: `C:\administrator\foo\test\types.js`
  - shows as: `C: istratorestres.js`  (<== so there's also a rogue `r` appearing here as well as many characters missing)

![image](https://user-images.githubusercontent.com/29628323/47295156-eaaaa580-d606-11e8-9219-f58147ab2257.png)

It looks like sometimes `\` in a file path escape the next character; but also, sometimes they escape half a dozen characters.

Here's a sample:

https://upload.clinicjs.org/public/3244631e31a1b6082c4b69dbf972bd2304e9ac23cd01dd48bca4bd1c804dff80/7044.clinic-flame.html

PR #193, opened 2018-10-24
Workaround for v8 tick processor bug

mafintosh:
Should fix https://github.com/davidmarkclements/0x/issues/192 until we can fix it in V8

### DIFF
--- a/lib/v8-log-to-ticks.js
+++ b/lib/v8-log-to-ticks.js
@@ -36,7 +36,7 @@ function escapePath (line) {
   const startPath = line.indexOf(':\\')
   if (startPath === -1) return line
   // \u005c is the unicode escape for the \ char
-  return line.slice(0, startPath) + line.slice(startPath).replace(/\\/g, '\\u005c')
+  return line.slice(0, startPath) + line.slice(startPath).replace(/([^\\])\\([^\\]|$)/g, '$1\\u005c$2')
 }
 
 // long lines make the --preprocess crash. We are just filtering them beforehand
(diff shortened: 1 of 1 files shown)


lib/v8-log-to-ticks.js
```diff
@@ -15,7 +15,28 @@ module.exports = wrapper
...
+  data = data.split('\n').map(escapePath).join('\n')
+  await writeFile(dest, data)
+  return dest
+}
+
+function escapePath (line) {
+  const startPath = line.indexOf(':\\')
+  if (startPath === -1) return line
+  // \u005c is the unicode escape for the \ char
+  return line.slice(0, startPath) + line.slice(startPath).replace(/\\/g, '\\u005c')
```
mcollina: If the V8 fix is in, would this code break?

AlanSl:
I've tested this on Windows 10 and it appears to fix https://github.com/davidmarkclements/0x/issues/192 so LGTM functionality-wise

lib/v8-log-to-ticks.js
mafintosh: Potentially yes, depending on how the fix lands (my current fork just disables all unescaping in their parser, which *would* break this).

Once that lands we have time to fix it here tho, so I think this is okay for now.

mafintosh: Actually only replacing [^\]\ with the escape sequence properly future proofs a lot better :+1:

AlanSl:
One observed issue which is maybe something for a separate PR - the names are coming out fine, but 0x is missclassifying some `app` and `deps` frames. For example from `0x/examples/rest-api` on Windows:

image

I'm going to do some more tests to see if it always misclassifies or if it's to do with particulars here such as the app code being in a subfolder and the dependency being installed outside of the project folder

AlanSl:
Probably. Let's release this and I'll post it as a separate issue.

LGTM

Issue #196, opened 2018-11-06
Ability to visualize profiling info from v8 profiler


slonka:
Hi,

V8 profiler has methods called StartProfiling, StopProfiling (https://v8docs.nodesource.com/node-10.6/d2/d34/classv8_1_1_cpu_profiler.html#ab78880137fc43d636cca23c81567b52d). There are modules that allow accessing these methods: https://github.com/hyj1991/v8-profiler-next/blob/master/src/cpu_profile.cc. I created a converter that can convert that format into something that is acceptable by 0x (generates a tree structure similar to the one generated here: https://github.com/davidmarkclements/0x/blob/master/lib/render.js#L19). 

Would you be interested in including that functionality? (you can preview the code here: https://github.com/slonka/0x/commit/a9c848ca9d2fa05577f02bd77230485959270fa9 and I'll create a PR if you want)

BTW I'm just wondering why did you only use `--prof` as a means of collecting stack information? How does it differ from `CpuProfiler::StartProfiling`?

mcollina:
> Would you be interested in including that functionality? (you can preview the code here: slonka@a9c848c and I'll create a PR if you want)

I think so! Can you add a screenshot here of how that would look like?

> BTW I'm just wondering why did you only use --prof as a means of collecting stack information? How does it differ from CpuProfiler::StartProfiling?

I _think_ because we need no native addon to start that. @davidmarkclements definitely knows more!

slonka:
> I think so! Can you add a screenshot here of how that would look like?

Yeah sure. Here is a flamegraph of a simple server responding with a string `hello` under 10RPS.
<img width="1531" alt="screen shot 2018-11-06 at 10 39 37" src="https://user-images.githubusercontent.com/2753650/48055882-e8248000-e1b0-11e8-853b-440894cd8394.png">

To run it you simply invoke: 
0x --visualize-v8-profile v8-profiler-output.json

> I think because we need no native addon to start that. @davidmarkclements definitely knows more!

I was more interested in the profiling data that is provided. Does `--prof` call `CpuProfiler::StartProfiling` underneath?

mcollina:
@slonka send the PR over. In your PR can you please add an example on how to generate that output?

> I was more interested in the profiling data that is provided. Does --prof call CpuProfiler::StartProfiling underneath?

I never investigated in depth. The actual output looks very similar anyway.

PR #197, opened 2018-11-06
Add support for visualizing v8 profile 

slonka:
...gathered by CpuProfiler::StartProfling.

As suggested in the comment: https://github.com/davidmarkclements/0x/issues/196#issuecomment-436196449 here is an example https://github.com/slonka/0x-visualize-v8-profile-example of how to generate v8 profile at runtime and a flamegraph from it.

Example flamegraph looks like this:
screen shot 2018-11-06 at 16 36 15

There is a couple of things that this PR is missing:
- [x] tests
- [x] merged / unmerged
- [x] optimized / unoptimized could probably be taken from "bailout reason"
- [ ] I do not know how 0x detects tiers

But the basic functionality is there. Would love help improving this.


lib/convert-v8-profile-to-tree.js
```diff
@@ -0,0 +1,28 @@
+function convert (profile) {
```
mcollina: Can you add `'use strict'`  at the top?

slonka:
@mcollina you just go into the example repo I linked to  https://github.com/slonka/0x-visualize-v8-profile-example, clone it, run `npm i` and `./test.sh` and you can see the `flamegraph.html` in the folder

mcollina:
Can you add that code in the https://github.com/davidmarkclements/0x/tree/master/examples and add a section about it in the README?

slonka:
Yeah sure, I thought that you had problems running the example.

readme.md
```diff
@@ -16,7 +16,7 @@ on any platform which Node runs on (macOs, Linux, Windows, Android...).
 * Node v8.5.0 and above
 * Default usage supports any Operating System that Node runs on!
 * Chrome
-  * Other browsers may open flamegraphs in a degraded, but functional form 
+  * Other browsers may open flamegraphs in a degraded, but functional form
```
slonka: I did not change it, some formatter did.

mcollina:
I think you can’t provide that info through that output. We can detect that by parsing the isolate.log file.

The buttons can either do nothing, or we can grey them out in this case.

slonka:
Yes, sure. How can I disable the "merge" button?

davidmarkclements:
if the v8-profiler doesn't differentiate between optimized/unoptimized we'll also need to disable the Optimized Unoptimized buttons as well

Additionally the filter buttons – possibly many of those don't apply? They should also be disabled


The buttons are HTML buttons so can have a disabled attribute set, this is passed down to a basic function component here: 

https://github.com/davidmarkclements/0x/blob/master/visualizer/cmp/controls.js#L3-L15

and they're called here: https://github.com/davidmarkclements/0x/blob/master/visualizer/cmp/controls.js#L19-L33


you might need to adjust/add to the state (https://github.com/davidmarkclements/0x/blob/master/visualizer/state.js) to determine whether a button should be disabled, 

the state is initialized in the visualizer here: https://github.com/davidmarkclements/0x/blob/master/visualizer/index.js#L52

and options can be passed to the visualizer from here: https://github.com/davidmarkclements/0x/blob/master/lib/render.js#L23

slonka:
@davidmarkclements - I disabled the buttons when flamegraph is generated from v8 profiler output.

It looks like this:
screen shot 2018-12-04 at 22 15 00

slonka:
BTW, this repo needs more tests definitely because the previous version broke the tool and I only noticed when I wanted to use it with different switches. I'll open a new PR for that.

mcollina:
@slonka some tests have been added in the last week. You might want to rebase.

slonka:
I did the rebase, any chance to get this merged? @mcollina @davidmarkclements

davidmarkclements:
amazing PR @slonka sorry it took so long, I'm just going to rename it to --visualize-cpu-profile – especially in light of https://github.com/nodejs/node/pull/27147

davidmarkclements:
renaming commit here:
https://github.com/davidmarkclements/0x/commit/d5c8d3b36bd59f7188be9c3655885a0b265489f7

slonka:
Thanks for merging this 🎊

slonka:
Are you sure the format of --cpu-prof matches the one provided here?

davidmarkclements:
No, however I expect it to be very similar to the extent that a tweak or two will make it work for both scenarios

PR #206, opened 2018-12-27
Disable default alphabetical sorting in flamegraph

sokra:
I figured the entries are in a weird order. Seem like d3-fg sorts nodes by default by name. This doesn't make sense for 0x as node are already in correct (time) order.


davidmarkclements:
Flame graphs are an aggregate of samples, there is no time axis (that’s flame charts). Alphabetically sorting is the norm (see http://www.brendangregg.com/flamegraphs.html)

PR #208, opened 2019-02-22
Skip redirecting stdout for Worker threads

addaleax:
This breaks running 0x with Workers, because the internal stdio
implementation for Workers currently relies on `process.stdout`
being the originally provided object.

### DIFF
--- a/lib/preload/redir-stdout.js
+++ b/lib/preload/redir-stdout.js
@@ -1,13 +1,21 @@
 'use strict'
 const net = require('net')
 
-const socket = new net.Socket({
-  fd: 3,
-  readable: false,
-  writable: true
-})
-Object.defineProperty(process, 'stdout', {
-  configurable: true,
-  enumerable: true,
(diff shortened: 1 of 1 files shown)


lib/preload/redir-stdout.js
```diff
@@ -1,6 +1,11 @@
 'use strict'
 const net = require('net')
 
+try {
+  // Skip redirecting stdout in Worker threads.
+  if (!require('worker_threads').isMainThread) return;
```
mcollina: Lint is failing. Would you mind adding an exception?

addaleax: @mcollina Ugh, just seeing that … since this is a parsing error for the linter, not a linter error, I’m not sure there’s a way to add an exception…?

I’ve worked around this the awkward way, wrapping the rest of the file in `if (!isWorker)`. I hope that’s okay?

mcollina:
Thanks!

PR #218, opened 2019-10-08
Flag likely wasm stack frames

goto-bus-stop:
wasm stack frames are outputted (output? put out?) basically with only their name, and no location data. I tried finding other differences between wasm and JS frames but there don't seem to be any.

This PR marks JS frames that don't have a source location as WASM frames. I'm not 100% sure that is right, but it should be close—Clinic Flame would've crashed in the past on JS frames without a filename and it has seen quite a few projects by now. Do we know if anything else may cause missing location data?

The [WASM] tag allows clinic flame to handle the wasm stack frames specially (eg. by setting a default filename or labelling them differently).

0x already "worked" without this, but wasm frames were categorised as V8 frames, which is quite inaccurate. This categorises wasm frames as "native" instead, which is also not great, but is at least not great in the same way as evals :)

### DIFF
--- a/lib/ticks-to-tree.js
+++ b/lib/ticks-to-tree.js
@@ -11,6 +11,10 @@ const nodeBootstrapRegExp = / internal\/bootstrap.+\.js:\d+:\d+$/
 
 module.exports = ticksToTree
 
+function hasFileLocation (jsFrameName) {
+  return /:\d+:\d+$/.test(jsFrameName)
+}
+
 function ticksToTree (ticks, options = {}) {
   const { mapFrames, inlined, pathToNodeBinary = process.execPath } = options
   const merged = {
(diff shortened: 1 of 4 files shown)


davidmarkclements:
Awesome - I think this might be worth having a separate classification for wasm with its own filter - I think some V8 stacks might have that same characteristic though - not sure - what V8 stacks are left after this change?

davidmarkclements:
Awesome - if all wasm frames have a double colon then maybe we can pattern match that 

I think kind is bogus here but it doesn’t matter

goto-bus-stop:
wasm frames for C functions like `memcpy` literally just have `"name": "memcpy"` unfortunately lol. the example I pasted has `::` everywhere because they are Rust stack frames, but Rust programs also use memcpy all over the place and have frames like `"name": "__rust_alloc"` and `"name": "__rust_dealloc"` for memory allocation. There's not much to go on :disappointed:

davidmarkclements:
Ohhhhh. Hmm. Ok. I think all of this can be improved with proffer integration but we need to update proffer to decode wasm stacks ( see https://github.com/davidmarkclements/proffer ) but for now let’s go with this approach. If you can add a filter (and a corresponding UI button?) this is good to go

goto-bus-stop:
Sample: https://4451070x-d1u0p7jg7.now.sh/flamegraph.html

davidmarkclements:
v4.9.1 released (extra patch for https://github.com/davidmarkclements/0x/commit/1db991a3ca9227642d290381bcc00257c34e6e7e)

Issue #129, opened 2018-06-02
Attaching 0x in the middle of the processing


behrad:
I want to start profiling with an interrupt or delay. Any options available?

mcollina:
in which operating system/node version?

It might be possible on Linux or Solaris using kernel tracing (perf/dtrace). I'm not sure if this is supported in `0x`.

behrad:
> in which operating system/node version?

CentOS 7, node v8.11.x

spalger:
I also would really benefit from being able to run this on a subset of a process, even if I can't attach to the process while it's running the ability to start and stop the collection within the lifecycle of the program would be awesome. Running on the latest macOS and node here.

davidmarkclements:
This is an advanced use case, you'll need attach perf or dtrace to the the process, generate stack output and then pass the stack output to 0x – all of this needs documenting though

mannypamintuan:
I am interested in this advanced use case.

Specifically, the ability to begin/start profiling with a function call (`start()`) and end profiling without the `soft exit`.
My specific use-case is to have a web app with  endpoints that allow the start and end of profiling to occur, retrieve the `.0x` files and generate flamegraphs locally.

I believe this is related to the original issue. (if not, I'd be happy to make separate :) )

@davidmarkclements 
I'd like to help on this if possible. 
I am relatively new to node.js and so if this issue is not suited for a `first time contributor` I'd love to help in any other way.

cc-ebay:
I would like to have this too. 
I would like to delay the profiling **after** a startup phase. 
It may be implemented listening for a signal, or a key or a programmatic function (ex: `require('0x').start()`

PR #222, opened 2020-02-13
Collect timeout feature

DylanC:
Putting this up early to have eyes on it. Node Clinic will have a new feature to set a timed delay before we collect the data. 

Relevant PR:
https://github.com/nearform/node-clinic/pull/208

For 0x the delay won't hold up the process but rather filter the data depending on the delay value passed in. Open to feedback and further changes/suggestions.

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -60,13 +60,13 @@ async function cmd (argv, banner = defaultBanner) {
       visualizeOnly: 'visualize-only',
       visualizeCpuProfile: 'visualize-cpu-profile',
       collectOnly: 'collect-only',
+      collectDelay: 'collect-delay',
       kernelTracing: 'kernel-tracing',
       kernelTracingDebug: 'kernel-tracing-debug',
       treeDebug: 'tree-debug',
       writeTicks: 'write-ticks',
       onPort: 'on-port',
-      P: 'onPort',
-      collectDelay: 'on-timeout'
+      P: 'onPort'
     }
   })
 
--- a/cmd.js
+++ b/cmd.js
@@ -65,7 +65,8 @@ async function cmd (argv, banner = defaultBanner) {
       treeDebug: 'tree-debug',
       writeTicks: 'write-ticks',
       onPort: 'on-port',
-      P: 'onPort'
+      P: 'onPort',
+      collectDelay: 'on-timeout'
     }
   })
 
--- a/lib/v8-log-to-ticks.js
+++ b/lib/v8-log-to-ticks.js
@@ -119,11 +119,16 @@ function v8LogToTicks (isolateLogPath, options) {
       close()
     })
 
+    const firstTick = [];
     const tickStream = parse('ticks.*', (tick) => {
       const addr = tick.s.filter((n, i) => i % 2 === 0)
       var stack = addr.map((n) => codes[n]).filter(Boolean)
       const delayMs = options.collectDelay * 1000
-      if (tick.tm > (tick.tm+delayMs)) {
+      if(firstTick.length===0) {
+        firstTick.push(tick.tm);
+      }
+      // Compare ticks to first for collectDelay
+      if (tick.tm > (firstTick[0]+delayMs)) {
         ticks.push(stack.reverse())
       }
     })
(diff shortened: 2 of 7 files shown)


cmd.js
```diff
@@ -65,7 +65,8 @@ async function cmd (argv, banner = defaultBanner) {
       treeDebug: 'tree-debug',
       writeTicks: 'write-ticks',
       onPort: 'on-port',
-      P: 'onPort'
+      P: 'onPort',
+      collectDelay: 'on-timeout'
```
goto-bus-stop:
```suggestion
      collectDelay: 'collect-delay'
```

lib/v8-log-to-ticks.js
```diff
@@ -69,7 +69,7 @@ function fixLines (line) {
 
 function v8LogToTicks (isolateLogPath, node) {
```
goto-bus-stop: let's rename this `node` parameter to something like `options`

cmd.js
DylanC: @goto-bus-stop - No problem. I'll update the arg in node-clinic to match.

lib/v8-log-to-ticks.js
DylanC: Indeed. Easier to understand.

goto-bus-stop:
(This functionality is tested in https://github.com/nearform/node-clinic-flame/pull/107. Flame is a bit better set up for testing atm :grimacing: )

@DylanC we need a line of documentation for this option in `docs/api.md`. I think there's also another call to the `v8LogToTicks()` function in index.js that needs to be updated for the new call signature.

DylanC:
@goto-bus-stop - I have both of those updates pushed into the branch.

lib/v8-log-to-ticks.js
```diff
@@ -122,7 +122,10 @@ function v8LogToTicks (isolateLogPath, node) {
     const tickStream = parse('ticks.*', (tick) => {
       const addr = tick.s.filter((n, i) => i % 2 === 0)
       var stack = addr.map((n) => codes[n]).filter(Boolean)
-      ticks.push(stack.reverse())
+      const delayMs = options.collectDelay * 1000
+      if (tick.tm > delayMs) {
```
goto-bus-stop: Can you try the comparison relative to the first tick's time here like we discussed in slack a while ago?

DylanC: I'll have a look at it!

lib/v8-log-to-ticks.js
```diff
@@ -122,7 +122,10 @@ function v8LogToTicks (isolateLogPath, node) {
     const tickStream = parse('ticks.*', (tick) => {
       const addr = tick.s.filter((n, i) => i % 2 === 0)
       var stack = addr.map((n) => codes[n]).filter(Boolean)
-      ticks.push(stack.reverse())
+      const delayMs = options.collectDelay * 1000
+      if (tick.tm > (tick.tm+delayMs)) {
```
goto-bus-stop: this will always evaluate to `false` :sweat_smile: The callback to `parse()` is called once for each tick. we need to remember the value for the first tick and compare to that in all subsequent ticks.

DylanC: Yeah I realized that pretty quick. :laughing: I've another commit up for it.

DylanC:
@goto-bus-stop - Should I wait until we figure out the tests for Flame first? Seems testing this would have the same issue?

ErisDS:
Wondering if there's anyone around to help me revive this PR?

I have a rebased version of this running here with added usage info + a minor tweak:

https://github.com/TryGhost/0x

It is working & useful - it sort of solves https://github.com/davidmarkclements/0x/issues/129 as although you can't explicitly say when to start profiling, you can skip the startup process of a script and only measure e.g. when firing requests.

I did look at adding the requested unit tests, but as far as I can tell the code that this affects in v8-log-to-ticks is currently untested, meaning adding a test for this behaviour is a fairly big task, and I don't really know where to start.

I don't really want to run a fork, ideally I'd like to get this populated through flame & clinic as was the intention of the original author.

mcollina:
would you like to do a fresh PR?

ErisDS:
Done:  https://github.com/davidmarkclements/0x/pull/239 :)

ErisDS:
Heads up there's nothing in here that wasn't in #239, and there for this can be closed 😺

PR #225, opened 2020-03-09
Add transpiled warning

DylanC:
Detect transpiled code by using a median of function name length and provide warning to the user. Also checks file for a source map comment.

### DIFF
--- a/index.js
+++ b/index.js
@@ -116,19 +116,16 @@ function checkForTranspiledCode (filename) {
   let isTranspiled = false
 
   // Check for a source map
-  isTranspiled = readFile.includes('//# sourceMappingURL=')
-
-  while ((matchedObj = regex.exec(readFile)) !== null) {
-    // Avoid infinite loops with zero-width matches
-    if (matchedObj.index === regex.lastIndex) {
-      regex.lastIndex++
-    }
+  if (readFile.includes('//# sourceMappingURL=')) {
+    isTranspiled = true
+  } else {
+    isTranspiled = true
     // Loop through results and check length of fn name
-    matchedObj.forEach((match, groupIndex) => {
-      if (groupIndex !== 0 && match.length < 3) {
-        isTranspiled = true
+    while ((matchedObj = regex.exec(readFile)) !== null) {
+      if (matchedObj[1].length > 3) {
+        isTranspiled = false
       }
-    })
+    }
   }
   return isTranspiled
 }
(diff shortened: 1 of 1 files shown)


mcollina:
I think https://github.com/davidmarkclements/0x/pull/185 might be also a good thing to add.

DylanC:
@mcollina After we add the warning I will work on the source map support next.

DylanC:
@mcollina - I have made this change compatible with node v8. Can we get this merged in?

index.js
```diff
@@ -25,6 +24,10 @@ async function zeroEks (args) {
     args.pathToNodeBinary = pathTo('node')
   }
 
+  if (checkForTranspiledCode(args.argv[0])) {
+    console.warn('Transpiled code is not supported')
```
goto-bus-stop: Should we use a similar message to Bubbleprof? 'The code is transpiled, 0x does not support source maps yet.'

I think this check should happen a bit later—right now it happens before any argument validation, and if your CLI arguments are wrong that is a more important issue to deal with. If you're using the `--visualize-only` flag, the input filename may not exist (or may not be provided in the first place).

I would probably put it just before the call to `startProcessAndCollectTraceData`, when we know for sure that we're actually starting a subprocess.

index.js
```diff
@@ -41,6 +40,11 @@ async function zeroEks (args) {
   if (visualizeCpuProfile) return cpuProfileVisualization(args)
 
   args.title = args.title || `node ${args.argv.join(' ')}`
+
+  if (checkForTranspiledCode(args.argv[0])) {
```
goto-bus-stop: Two more edge cases:
- `args.argv` may be empty if no file path is given:
  ```
  ./cmd.js --
  ```
  In that case I think we should skip the check.
- The file may not exist: it's better if we let the Node.js process start and give an error for that instead of throwing an error from 0x internals that users may not be familiar with
  ```
  ./cmd.js -- node doesnotexist.js
  ```
  So if `readFileSync()` throws an error, we should ignore that and keep going.

DylanC: Check the latest changes and let me know if its good.

index.js
```diff
@@ -41,6 +40,11 @@ async function zeroEks (args) {
   if (visualizeCpuProfile) return cpuProfileVisualization(args)
 
   args.title = args.title || `node ${args.argv.join(' ')}`
+
+  if (args.argv && checkForTranspiledCode(args.argv[0])) {
```
goto-bus-stop: An empty array also evaluates to `true`. A reliable way to check that `argv` is an array with at least 1 element is:
```js
Array.isArray(args.argv) && args.argv.length > 0
```

DylanC:
@mcollina - Can we get this merged?

PR #227, opened 2020-03-16
Source map support

DylanC:
After we merge #225 it would be nice to add support for source map usage. I'm not sure what is the best approach but I was thinking of continuing the work from #185. This branch includes the previous commits of the original PR. (Rebased on master)


DylanC:
@davidmarkclements or @mcollina - Do you have any input as to what else is needed here?

lib/ticks-to-tree.js
```diff
@@ -15,19 +19,22 @@ function hasFileLocation (jsFrameName) {
...
-    value: 0,
-    top: 0,
-    children: []
-  }
-  const unmerged = {
-    name: 'all stacks',
-    value: 0,
-    top: 0,
-    children: []
+async function ticksToTree (ticks, options = {}) {
```
mcollina: Is it necessary to turn this into an `async` function?

lib/ticks-to-tree.js
```diff
@@ -99,71 +112,184 @@ function ticksToTree (ticks, options = {}) {
...
-
-      if (ix === stack.length - 1) frame.top++
-      if (ix === 0) merged.value += 1
-      frame.value++
-
-      lastFrame = frame
+  if (sourceMaps) {
+    await Promise.all(Object.keys(jsLibraries).map(async (jsLibrary) => {
+      return loadSourceMap(jsLibrary, sourceMaps)
+    }))
```
mcollina: This is going to be pretty expensive, this should use p-limit or p-map.

lib/ticks-to-tree.js
```diff
@@ -99,71 +112,184 @@ function ticksToTree (ticks, options = {}) {
...
+  Object.keys(loadedSourceMaps).forEach((file) => {
+    if (loadedSourceMaps[file] !== false) {
+      loadedSourceMaps[file].destroy()
+    }
+  })
+
+  return trees
+}
+
+const loadedSourceMaps = {}
```
mcollina: This should not be a global. It creates a memory leak

lib/ticks-to-tree.js
DylanC: Not sure. I'm picking up this from previous PR changes. I don't see why it is like this currently.

lib/ticks-to-tree.js
DylanC: Will do!

lib/ticks-to-tree.js
DylanC: No problem. I'll change this.

lib/ticks-to-tree.js
DylanC: Ah, it uses await here:
`if (sourceMaps) {
    await Promise.all(Object.keys(jsLibraries).map(async (jsLibrary) => {
      return loadSourceMap(jsLibrary, sourceMaps)
    }))
  }`
I don't have an answer to your question really. I would likely need the original author to chime in. @pimlie ?

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,184 @@ function ticksToTree (ticks, options = {}) {
...
-
-      if (ix === stack.length - 1) frame.top++
-      if (ix === 0) merged.value += 1
-      frame.value++
-
-      lastFrame = frame
+  if (sourceMaps) {
+    await pMap(Object.keys(jsLibraries).map(async (jsLibrary) => {
+      return loadSourceMap(jsLibrary, sourceMaps)
+    }))
```
mcollina: Please add a concurrency of 16, check the p-map doc.

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,184 @@ function ticksToTree (ticks, options = {}) {
...
-      if (child === undefined) children.push(frame)
-      else frame = child
-
-      if (ix === stack.length - 1) frame.top++
-      if (ix === 0) merged.value += 1
-      frame.value++
-
-      lastFrame = frame
+  if (sourceMaps) {
+    await pMap(Object.keys(jsLibraries).map(async (jsLibrary) => {
```
mcollina: This is incorrect, the mapper is a the second argument. Check the p-map docs.

lib/ticks-to-tree.js
DylanC: @mcollina Do you have any suggestions for this? loadedSourceMaps is used in 3 different functions.

lib/ticks-to-tree.js
DylanC: I actually had it the other way around first but the linted complained that pMap wasn't being used.

lib/ticks-to-tree.js
mcollina: You need to pass it as a parameter.

DylanC:
@mcollina - I've changed loadedSourceMaps to not be global any longer. Anything else required here?

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
+    }
+
+    for (let i = 0; i < sourceMapFiles.length; i++) {
+      if (sourceMapFiles[i]) {
+        // convert relative source map paths to absolute paths
+        if (!isAbsolute(sourceMapFiles[i])) {
+          sourceMapFiles[i] = resolve(dirname(fileName), sourceMapFiles[i])
+        }
+
+        if (existsSync(sourceMapFiles[i])) {
```
mcollina: Please use an async version of this.

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
+          try {
+            debug(`loading source map from ${sourceMapFiles[i]}`)
+            const sourceMap = JSON.parse(readFileSync(sourceMapFiles[i]))
```
mcollina: please read the file asynchronously

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
+            loadedSourceMaps[fileName] = await new SourceMapConsumer(sourceMap)
+            if (!loadedSourceMaps[fileName].sourceRoot) {
+              loadedSourceMaps[fileName].sourceRoot = dirname(sourceMapFiles[i])
+            }
+            break
+          } catch (e) {
+            console.error(e)
```
mcollina: Are we sure this is correct? If we can't read the file, is it safe to just carry on?

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
-        else if (frame.S === 2) frame.name = '*' + frame.name
-        else if (frame.S === 3) frame.name = '~' + frame.name + ' [INLINABLE]'
-        else if (frame.S === 4) frame.name = '*' + frame.name + ' [INLINABLE]'
-        children.push(frame)
-      } else frame = child
+      if (source.source) {
+        name = `${key[0]} ${source.source}:${source.line}:${source.column}`
+      }
+    } catch (err) {
+      console.error(err, fileName, line, column)
```
mcollina: Are we sure this is correct? If we can't read the file, is it safe to just carry on?

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
+
+      // only rename when fileName is an absolute path
+      if (isAbsolute(fileName)) {
+        const relativeName = relative(options.relativePath, fileName)
+        frame.name = `${key[0]} ${relativeName}:${line}:${column}`
+      }
+    }
+  }
+  if (frame.children && frame.children.length) {
+    frame.children.forEach((child) => renameFrame(options, child, loadedSourceMaps))
```
mcollina: Please use a for loop, it'll be faster.

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
   }
+  delete frame.type
```
mcollina: can we just set type to undefined? It'll be faster.

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
+}
+
+function addToMergedTree (trees, stack) {
+  var lastFrame = null
+  stack.forEach((frame, ix) => {
```
mcollina: Please use a for(;;) loop.

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,187 @@ function ticksToTree (ticks, options = {}) {
...
+    if (ix === 0) trees.merged.value += 1
+    frame.value++
+
+    lastFrame = frame
+  })
+}
+
+function addToUnmergedTree (trees, stack) {
+  var lastFrame = null
+  stack.forEach((frame, ix) => {
```
mcollina: Please use a for(;;) loop

lib/ticks-to-tree.js
DylanC: I was going to use fs.exists but the linter warned this is deprecated. Should I use fs.access instead?

DylanC: Using fs.access instead. :+1:

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,191 @@ function ticksToTree (ticks, options = {}) {
...
+  let loadedSourceMaps = {}
+  if (sourceMaps) {
+    const mapper = async (jsLibrary) => {
+      loadedSourceMaps = loadSourceMap(jsLibrary, sourceMaps, loadedSourceMaps)
+      return loadedSourceMaps
+    }
+    await pMap(Object.keys(jsLibraries), mapper, { concurrency: 16 })
+  }
+  if (sourceMaps || relativePath) {
+    Object.keys(trees).map((tree) => {
```
mcollina: this does not need to be a map, just a forEach. I would replace it with a for loop instead.

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,191 @@ function ticksToTree (ticks, options = {}) {
...
+      trees[tree].children.map((frame, index) => {
```
mcollina: same, this can be a for loop

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,191 @@ function ticksToTree (ticks, options = {}) {
...
+        renameFrame(options, frame, loadedSourceMaps)
+      })
     })
   }
+  Object.keys(loadedSourceMaps).forEach((file) => {
+    if (loadedSourceMaps[file] !== false) {
+      loadedSourceMaps[file].destroy()
+    }
+  })
```
mcollina: Use a for loop

lib/ticks-to-tree.js
```diff
@@ -99,71 +113,191 @@ function ticksToTree (ticks, options = {}) {
...
+    if (typeof sourceMaps === 'object') {
+      if (sourceMaps[basename(fileName)]) {
+        sourceMapFiles.unshift(sourceMaps[basename(fileName)])
+      }
+      if (sourceMaps[fileName]) {
+        sourceMapFiles.unshift(sourceMaps[fileName])
+      }
+    }
+
+    if (access(fileName)) {
```
mcollina: are you sure this is sync?

lib/ticks-to-tree.js
mcollina: Is it actually safe?

lib/ticks-to-tree.js
DylanC: Oh its not actually. Updated now!

DylanC:
> Does this include some unit tests for the transpiled case?

This will need to include some unit tests. An example exists but not a unit test.

mcollina:
CI seems to be failing

DylanC:
@mcollina Yep.

Its failing on this:
`not ok 2 - ENOENT: no such file or directory, access '[eval]'`

Line 167:
`if (accessSync(fileName)) {`

I'm looking into the flow of this feature at the moment. Something is not right here with the source map code.

Issue #228, opened 2020-03-16
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined


Jero786:
Hello!

Thank for your library, it's such amazing. 
The only problem that I'm facing is about I'm not able to define an absolute path different than `__dirname`.

```js
async function captureFlameGraph (timestamp) {
    const fileName = '/Users/jero/projects/ipsy/shopper-web-3/server/dumps'
    const opts = {
        argv: [path.join(__dirname, 'index.js'), '--output-dir' ,`"/some_path_here_doesnt_work"`], // <---- seting output-dir it doesn't effect nothing at all.
        workingDir: __dirname, // <------- That path it's working 
        workingDir: // '/other_absolute_path/' <------ Throw exception, please take at the bottom*
    };
    try {
        const file = await zeroEks(opts);
    } catch (e) {
        console.error(e)
    }
}
```

Debugging internally seems like the absolute path that we define, it's not creating the `isolate-..-v8.log`, so when check if exit throws en exception in the following line: https://github.com/davidmarkclements/0x/blob/master/platform/v8.js#L110, after that, arrive in the line: https://github.com/davidmarkclements/0x/blob/master/index.js#L83 and pass the second argument as a `undefined`, because it didn't found isolate log in the workingDir. And finally, in the join method, we have the `undefined` unwanted.

![image](https://user-images.githubusercontent.com/1815487/76801804-1f4b6e00-67b5-11ea-9599-84524b7ddb99.png)

That produces the error :(

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined
    at validateString (internal/validators.js:125:11)
    at join (path.js:1147:7)
    at v8.catch (/Users/jero/projects/ipsy/shopper-web-3/node_modules/0x/index.js:85:27)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```
Anyone could help me on that?

Jero786:
Here I mead the fix of this issue here: https://github.com/davidmarkclements/0x/pull/229

divmgl:
So is this library completely broken? I can't get a simple use-case to run from the `README.md`.

SukkaW:
> So is this library completely broken? I can't get a simple use-case to run from the `README.md`.

Yes.

PR #229, opened 2020-03-16
fix:issue:228 - TypeError [ERR_INVALID_ARG_TYPE]: The path argument m…

Jero786:
# Fix: issue #228

"TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined".

Please for more detail, take a look at the open issue: #228

### DIFF
--- a/platform/v8.js
+++ b/platform/v8.js
@@ -25,7 +25,7 @@ async function v8 (args) {
 
   let proc = spawn(pathToNodeBinary, [
     '--prof',
-    `--logfile=%p-v8.log`,
+    `--logfile=${workingDir ? `${args.workingDir}/` : ''}%p-v8.log`,
     '--print-opt-source',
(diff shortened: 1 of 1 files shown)


mcollina:
Could you add a test?

SukkaW:
@Jero786 Would you mind catching up with the PR then?

@mcollina So you guys just leave the issue unsolved just because the PR here is hanged?

cdbattags:
Just wanted to bump. Waiting on this as well.

SukkaW:
@mcollina Since @Jero786 hasn't been active for a long time. Will you implement a fix yourself? Or at least tell people (who want to contribute) which unit test case to be updated. Because, come on, the programmatic API of the library is **COMPLETELY BROKEN** and which means the unit test cases is not reliable AT ALL.

Jero786:
Sorry Guys, I don't have time to do the UT. I just wanted to contribute to what I did to resolve the issue that I faced.

PR #230, opened 2020-08-04
fix regex bug of optional 0x prefix

gabylb:
The bug prevented a file match on a system in which %p or std::ostream&<< doesn't prefix the address with 0x, as on z/OS.

Note the updated regex now also matches the 0x part in index.js visualize().


mcollina (approved):
lgtm

PR #232, opened 2020-09-24
fix: error msg on workingdir

ovhemert:
When using the API, and specifying a different `workingDir`, the process crashes with the following error:

`TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined`

It was originally reported in issue #228 and and fixed in PR #229 but a test is still missing to get it merged.
This PR applies the same fix, and adds a test to validate the error does not occur anymore.

### DIFF
--- a/platform/v8.js
+++ b/platform/v8.js
@@ -25,7 +25,7 @@ async function v8 (args) {
 
   let proc = spawn(pathToNodeBinary, [
     '--prof',
-    `--logfile=%p-v8.log`,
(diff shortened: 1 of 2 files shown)


ovhemert:
seems to fail on CI somehow => WIP

mcollina:
Is this still a wip?

ovhemert:
Nope... working now. Forgot to remove WIP i guess

mcollina (approved):
lgtm

PR #239, opened 2021-08-17
Collect timeout feature

ErisDS:
Original PR is here: https://github.com/davidmarkclements/0x/pull/222

> For 0x the delay won't hold up the process but rather filter the data depending on the delay value passed in. Open to feedback and further changes/suggestions.

This as the effect of only rendering flamegraphs with data after the delay ends. 

This solves some use cases of https://github.com/davidmarkclements/0x/issues/129. 
My own use case is not wanting to see output from the initial boot of my app (different problem, distracting!) but rather focus on what happens when requests are received. 

I am aware the code is untested, keen for some input on where to start with adding tests as the critical affected code in  lib/v8-log-to-ticks.js appears to be as yet untested as far as I can tell.

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -60,13 +60,13 @@ async function cmd (argv, banner = defaultBanner) {
       visualizeOnly: 'visualize-only',
       visualizeCpuProfile: 'visualize-cpu-profile',
       collectOnly: 'collect-only',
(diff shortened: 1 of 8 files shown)


mcollina:
Thanks!

ErisDS:
Thank you ❤️

PR #250, opened 2022-01-21
Adding the SIGTERM to soft exit

arthurguedes375:
When I was using  docker-compose I could not generate the flame-graph because the default signal sent by docker-compose is SIGTERM, there are ways to send SIGINT instead of SIGTERM but that adds more complexity to the docker-compose file, and since SIGTERM is not a force exit signal I think that it should be handled the same way that SIGINT is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "0x",
-  "version": "5.0.0",
+  "version": "5.0.1",
   "description": "🔥 single-command flamegraph profiling 🔥",
   "main": "index.js",
   "bin": "./cmd.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "0x",
-  "version": "5.0.0",
+  "version": "5.0.1",
```
mcollina, line 3: Can you revert this? We'll bump it on releaze

mcollina (commented):
Good work! Could you document this in the README?

package.json
arthurguedes375: Sure, i got it!

RafaelGSS (approved):
LGTM.

PR #252, opened 2022-01-26
fix: add support to inlined functions + sparkplug on v16+

RafaelGSS:
I'm still looking if there is something missing because looks like the call stack are not completely correct. 

By the way, I'm running this fork: https://github.com/RafaelGSS/slow-rest-api

```console
node ../node-clinic/bin.js flame --on-port='npm run test' -- node index.js
```

and I'm getting the graph attached

image

Look that `payload` function is side by side of `a`/`b`/`c`/ function which is its caller

image

Does that is acceptable? @mcollina

[204430.clinic-flame.zip](https://github.com/davidmarkclements/0x/files/7945683/204430.clinic-flame.zip)

### DIFF
--- a/lib/ticks-to-tree.js
+++ b/lib/ticks-to-tree.js
@@ -6,8 +6,7 @@ const debug = require('debug')('0x: ticks-to-tree')
 const escape = require('escape-string-regexp')
 
 const preloadDirRx = RegExp(join(escape(__dirname), 'preload'))
-// TODO(@rafaelgss): it looks like not working with v16 when node prefix was added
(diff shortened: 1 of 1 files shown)


RafaelGSS:
Well, looks like it still happens even before my fix (which is unrelated to that). So, it's ready to go.

mcollina (approved):
lgtm

RafaelGSS:
I did a second deeper look at it and looks good. Unfortunately, `inlined` functions are usually skipped from call-stack: https://github.com/davidmarkclements/0x/blob/master/lib/ticks-to-tree.js#L118

but this is a different issue, I think that it could land. Would you mind merging?

mcollina:
should I release or would you like to fix the other problem too?

PR #253, opened 2022-02-15
fix: collect-only to render-only flow

adiktofsugar:
In v8-log-to-ticks.js:152
```
const delayMs = options.collectDelay * 1000
...
// Compare ticks to first for collectDelay
if (tick.tm > (firstTick[0] + delayMs)) {
  ticks.push(stack.reverse())
}
```
But without this change, `collectDelay` is `NaN` so `firstTick[0] + NaN` ends up `NaN`, which always returns false when compared to a number, so this never resolves to true, and no ticks get added.

### DIFF
--- a/index.js
+++ b/index.js
@@ -123,12 +123,12 @@ async function cpuProfileVisualization (opts) {
   return file
 }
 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #254, opened 2022-02-17
fix: call soft close instead of process.exit when CTRL + C is pressed

rubenpad:
0x is not generating the flame graph when `CTRL + C` is pressed.

It is due to the process end up having two listeners for the `SIGINT` signal `[process.exit, softClose]`.

This could also fix the issue on https://github.com/clinicjs/node-clinic-flame/issues/161

### DIFF
--- a/lib/preload/soft-exit.js
+++ b/lib/preload/soft-exit.js
@@ -1,9 +1,4 @@
 'use strict'
 
-const SOFT_EXIT_SIGNALS = ['SIGINT', 'SIGTERM']
-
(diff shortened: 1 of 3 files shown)


RafaelGSS (changes requested):
I don't understand why it solves the issue, looks like you are doing the same thing. Can you elaborate?

RafaelGSS (approved):
LGTM

mcollina (approved):
lgtm

PR #255, opened 2022-03-09
fix: override process.stdout.fd

RafaelGSS:
Address: https://github.com/clinicjs/node-clinic/issues/337

It's still WIP because I'm looking to other approaches, the current one looks quite hacky to me.


mcollina (approved):
lgtm - this is as hacky as the code that introduced the problem.

mcollina (approved):
lgtm

Issue #215, opened 2019-07-31
Profile from Chromium 75 not loading with --visualize-cpu-profile


cmdcolin:
I used the "Javascript profiling" tab in Chromium 75 on linux

```
🔥  Creating flamegraph from v8 profile CPU-20190730T212046.cpuprofileTypeError:
🚫  Cannot read property 'children' of undefined
    at convert (/usr/lib/node_modules/0x/lib/cpu-profile-to-tree.js:5:33)
    at cpuProfileToTree (/usr/lib/node_modules/0x/lib/cpu-profile-to-tree.js:25:21)
    at render (/usr/lib/node_modules/0x/lib/render.js:26:13)

```

The resulting file failed to load. I am uploading it here just for reference

[CPU-20190730T212046.cpuprofile.gz](https://github.com/davidmarkclements/0x/files/3449383/CPU-20190730T212046.cpuprofile.gz)

The format of the profile format always seems to change at chromes whim unfortunately.

davidmarkclements:
did you gzip it yourself or did it come that way? did you try gunzipping it?

cmdcolin:
I gzipped it for upload. There might be a difference between this cpuprofile and the sample cpuprofile in this repo. I did obtain it from the "Javascript profiling" tab of Chromium

![2019-07-30-211726_3840x2160_scrot](https://user-images.githubusercontent.com/6511937/62221527-c642dd00-b37f-11e9-9e66-c5fc45474a5b.png)

The samples.cpuprofile in this repo can also be loaded into https://thlorenz.com/flamegraph/web/ but it appears similarly to have changed and was tracked here https://github.com/thlorenz/flamegraph/issues/21

cmdcolin:
If this is very similar to https://github.com/thlorenz/flamegraph/issues/21 that this is probably an enhancement type issue rather than a bug (but it would be very cool to have)

cmdcolin:
Found a tool that can work with latest chrome profiling JSON http://jamie-wong.com/post/speedscope/

PR #256, opened 2022-06-08
Support `.cpuprofile` of `node --cpu-prof`

kazarmy:
This pr adds support for the `.cpuprofile` file generated by `node --cpu-prof`, which is actually slightly different in format compared to the one generated by `v8-profiler-next` (an example is provided [here](https://raw.githubusercontent.com/kazarmy/0x/93b1d37ffa31d7e68282b6f46165da8c05bfd7f8/examples/cpu-profiler/samples-cpu-prof.cpuprofile) as part of the pr). An example of `node --cpu-prof` invocation has been added as part of `examples/cpu-profiler/test.sh` where it is used to profile `autocannon`.


kazarmy:
Fixes #215.

lib/cpu-profile-to-tree.js
```diff
@@ -21,8 +21,44 @@ function convert (profile) {
...
+  for (let i = 0; i < childrenLength; i++) {
+    converted.children[i] = convertCpuProf(nodes, id2Index, node.children[i])
+    converted.value += converted.children[i].value
+  }
+  return converted
+}
+
 function cpuProfileToTree (profile) {
-  const converted = convert(profile.head)
+  const converted = profile.head ?
```
RafaelGSS, line 59: Are you sure the unique way to identify both profiles is through the `head` property?

examples/cpu-profiler/test.sh
```diff
@@ -7,11 +7,13 @@ sleep 1
 
 curl -v http://localhost:3000/start-profiling
 
-node node_modules/.bin/autocannon -l -R 200000 -c 2000 -d 20 http://localhost:3000
+node --cpu-prof --cpu-prof-name=samples-cpu-prof.cpuprofile \
```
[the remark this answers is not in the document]
kazarmy: If you mean profiling `node index.js`, I tried that but couldn't get its `.cpuprofile` to be generated reliably, possibly because of the `kill -KILL`. After trying `kill -TERM`, `kill -INT` and just plain `kill` but still no joy, I settled for autocannon because its `.cpuprofile` isn't much bigger than the current `samples.cpuprofile` produced by `index.js`.

(I also tried profiling 0x ... its `.cpuprofile` is about 10 times bigger than the one for autocannon)

lib/cpu-profile-to-tree.js
kazarmy: It's not unique ... one can also check for `profile.nodes` (the screenshot at https://github.com/davidmarkclements/0x/issues/215#issuecomment-516880388 suggests other possible keys) but I chose perhaps the simplest way by basing the check on current code.

examples/cpu-profiler/test.sh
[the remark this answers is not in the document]
RafaelGSS: I would add another cpu-profiler test instead of incrementing this one. Something like `examples/cpu-profiler/cpu-prof-test.sh` and testing a HTTP application should be enough

lib/cpu-profile-to-tree.js
RafaelGSS: I'm just wondering if this is safe to use. I mean, V8 can decide to change the top structure, and it will silently fail

examples/cpu-profiler/test.sh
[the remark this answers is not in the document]
kazarmy: Ok 7698b50084b2b5088f181aee4620ad6351f08cd7

lib/cpu-profile-to-tree.js
kazarmy: Major changes to the structure would probably result in a stack trace similar to what is inside https://github.com/davidmarkclements/0x/issues/215#issue-474890398. Tbh I don't see how V8 can reasonably naturally evolve the top structure while also causing silent failure.

Bugs in 0x itself are outside the scope of this pr.

mcollina (approved):
lgtm

kazarmy:
merge?

mcollina:
Seems the linting is failing: https://github.com/davidmarkclements/0x/runs/6844137924?check_suite_focus=true#step:5:16

kazarmy:
Ok fixed 6f3aefce32f43f6537074b89831249454f7f9dd4. I've also now running the linter locally so there shouldn't be any linter errors after this.

kazarmy:
merge now?

kazarmy:
thx!

PR #257, opened 2022-06-16
Add test for #256 (`--cpu-prof` `.cpuprofile`)

kazarmy:
This pr adds a test for #256, which currently has no tests.


mcollina (approved):
lgtm

PR #258, opened 2022-06-18
cpu-profile-to-tree: Refactor out the filtering of parenthesized nodes

kazarmy:
This pr refactors out the code that does the filtering of parenthesized nodes like `(idle)` and `(garbage collector)` in `cpu-profile-to-tree.js`, as well as doing some related cleanup and testing.

As a side note, it appears that `(program)` [refers to V8 infrastructure code](https://github.com/nodejs/node/blob/61c7103ed06e5364eddae4586f90ee97531e14a6/deps/v8/src/profiler/symbolizer.cc#L28-L38) and that `(idle)`, `(garbage collector)` and `(program)` are guaranteed [not to have any children](https://github.com/nodejs/node/blob/61c7103ed06e5364eddae4586f90ee97531e14a6/deps/v8/src/profiler/symbolizer.cc#L172-L181) but this does assume that I'm reading the code correctly and that conclusions made on Node's V8 code is generalizable to other V8 implementations.

### DIFF
--- a/lib/cpu-profile-to-tree.js
+++ b/lib/cpu-profile-to-tree.js
@@ -1,20 +1,27 @@
 'use strict'
 
-function convert (profile) {
+// Filter out all parenthesized nodes with no children i.e. (idle), (garbage
+// collector) and (program)
+function filterParenthesizedNoChild (profileNode, convertedNode) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #273, opened 2023-07-25
Set black background on dark theme

allevo:
Enabling the dark mode if it is configured at the OS level.

This is how it will look like
image


RafaelGSS (approved):

mcollina (approved):
lgtm

PR #276, opened 2023-08-30
Allow stdin to work while profiling

### LINKED ISSUE #275: When running the command it kills the child process.
```
~/soundxyz/code/tools/admin-tools on jake/transfer-event-simpler*
$     TZ=UTC IS_ADMIN_TOOLS=1 0x --collect-only -- node --loader=bob-tsm src/index.ts
🔥  Profiling(node:1573822) ExperimentalWarning: Custom ESM Loaders is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
? Choose an environment (Use arrow keys)
  development 
  preview 
❯ staging 
🔥  Stats collected in folder file:///home/raynos/soundxyz/code/tools/admin-tools/1573822.0x
raynos at system76-pc  
```

I expected it to profile the child process and generate an artifact that can be turned into a flamegraph

My child process is waiting on `stdin` response for me to choose an option but instead `0x` killed it. 

Can `0x` pass the stdin through instead so I can select an option ?

STRd6:
Fixes #275

### DIFF
--- a/platform/v8.js
+++ b/platform/v8.js
@@ -33,7 +33,7 @@ async function v8 (args) {
     '-r', path.join(__dirname, '..', 'lib', 'preload', 'soft-exit'),
     ...(onPort ? ['-r', path.join(__dirname, '..', 'lib', 'preload', 'detect-port.js')] : [])
   ].concat(args.argv), {
-    stdio: ['ignore', 'pipe', 'inherit', 'pipe', 'ignore', 'pipe']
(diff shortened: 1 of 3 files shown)


mcollina:
Could you add a test for this?

mcollina (approved):
lgtm

RafaelGSS (approved):

PR #280, opened 2024-10-23
fix: node v20 complains of invalid arguments

eliOcs:
0x didn't work for me when using node v20 because calling process.exit with invalid arguments will throw an exception


mcollina (approved):
lgtm

PR #291, opened 2025-07-07
Allow configuring the cwd of the profiler via cli flags

### LINKED ISSUE #290: Support overriding the default workingDir to another path
When running a nodejs process in a restricted environment (eg. read only storage/container, user without write access, sandboxing) the 0x CLI is causing profiling logs [1] to be redirected to stdout, causing a flood in the logs. I assume this happens when the profiler fails to write to the `args.workingdir` as in our case.

This can be avoided if for example there is a configurable path where the process can have write access (eg `/tmp`).

In our case writing a custom wrapper around the `0x` API that also passes a workingdir with write access solved our issues.

https://github.com/davidmarkclements/0x/blob/a07a3a8c05c8238711130ecd588b932493e5ae36/platform/v8.js#L217

In case that's something that you are interested to support I am happy to send patches.

johngian:
Fixes #290

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -52,6 +52,7 @@ async function cmd (argv, banner = defaultBanner) {
       quiet: 'q',
       open: 'o',
       'output-dir': 'outputDir',
+      'working-dir': 'workingDir',
       D: 'outputDir',
       'output-html': 'outputHtml',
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

